PR #11, opened 2015-06-30
#29865 Added panel support

manneohrstrom:
This adds panel support to Nuke by adding the following two public methods:

```
engine.show_panel(title, bundle, widget_class)
engine.register_panel(title, bundle, widget_class)
```

The panel supports handles the case where a panel is requested to persist across sessions (e.g. a saved UI layout). The interface is likely to be pushed up to core and be made a general engine feature.

Example of an active panel based app:

image


engine.py
```diff
@@ -103,7 +104,16 @@ def init_engine(self):
...
-            
+        
+    
+    def pre_app_init(self):
+        """
+        Called at startup, but after QT has been initialized
+        """
+        # note! not using the import as this confuses nuke's calback system
+        # (several of the key scene callbacks are in the main init file...)
+        import tk_nuke
```
jfboismenu-adsk: Won't this make a certain part of the engine not reloadable anymore tough?

engine.py
```diff
@@ -197,6 +211,153 @@ def log_error(self, msg):
...
+                             mustn't take any arguments.        
+        """
+        # generate unique identifier
+        panel_id = self._generate_panel_id(title, bundle)
+        
+        # track registered panels
+        self.log_debug("Registering panel '%s'" % panel_id)
+        self._panels[panel_id] = {"title": title, "bundle": bundle, "widget_class": widget_class}
+
+        # tell nuke how to create this panel
```
jfboismenu-adsk: What will happen if you relaunch Nuke without Toolkit?

python/tk_nuke/panels.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2013 Shotgun Software Inc.
```
jfboismenu-adsk: 2015

engine.py
manneohrstrom: correct - the nuke engine was never reloadable (there were other `import tk_nuke` in the code already). I don't remember the details, but it was to do with how the nuke callbacks work. Engine reload is definitely less critical than app reload :)

engine.py
manneohrstrom: Glad you asked :)

Nuke will issue a warning and you are left with a blank panel. Would be cool if we could do somehing nicer than just a blank panel, but i think our options are pretty limited!

manneohrstrom:
hey @jfboismenu-adsk This is ready for another round of code review! Sorry it took a while :)

manneohrstrom:
@robblau FYI please note that i also updated the icons as part of this pull request so that the nuke engine is no longer using the TK icon:

image

engine.py
```diff
@@ -197,6 +211,153 @@ def log_error(self, msg):
...
+        Given a dialog name and a bundle, generate a Nuke panel id.
+        This panel id is used by Nuke to identify and persist the panel.
+        
+        This will return something like 'shotgun_tk_multi_loader2_main'
+        
+        :param dialog_name: An identifier string to identify the dialog to be hosted by the panel
+        :param bundle: The bundle (e.g. app) object to be associated with the panel
+        :returns: a unique identifier string 
+        """
+        panel_id = "shotgun_%s_%s" % (bundle.name, dialog_name)
```
jfboismenu-adsk: Feels like something that should be in core if the implementation is always the same.

engine.py
```diff
@@ -197,6 +211,153 @@ def log_error(self, msg):
...
+        panel_id = re.sub("\W", "_", panel_id)
+        panel_id = panel_id.lower()
+        self.log_debug("Unique panel id for %s %s -> %s" % (bundle, dialog_name, panel_id))
+        return panel_id 
+        
+    def _panel_factory_callback(self, panel_id):
+        """
+        Given a panel id, generate a panel UI.
+        
+        This method is intended to be executed by nuke itself as a callback
```
jfboismenu-adsk: Seems like I was right ;)
https://github.com/shotgunsoftware/tk-maya/pull/7/files#r34160136

engine.py
```diff
@@ -197,6 +211,153 @@ def log_error(self, msg):
...
+            # so we can assume that our _panels lookup data structure has a matching
+            # object present
+            panel = self._panels[panel_id]
+            # create a new panel widget
+            panel_widget = tk_nuke.NukePanelWidget(panel["title"], panel_id, panel["widget_class"])
+            # and add it to the current pane context (nuke handles this state)
+            panel_widget.addToPane()
+            
+        except Exception, e:
+            self.log_exception("Could not generate panel UI for panel id '%s'" % panel_id)
```
jfboismenu-adsk: We [are](https://github.com/shotgunsoftware/tk-maya/pull/7/files#r34182112) checking for engine not being ```None```in the maya engine but not here. Is this a bug? Another argument in favor of refactoring this logic in a base class.

engine.py
```diff
@@ -197,6 +210,186 @@ def log_error(self, msg):
...
+                                                   panel["widget_class"],
+                                                   *panel["args"],
+                                                   **panel["kwargs"])
+            # and add it to the current pane context (nuke handles this state)
+            panel_widget.addToPane()
+            
+            return panel_widget
+            
+        except Exception, e:
+            # catch-em-all here because otherwise Nuke will just silently swallow them
```
jfboismenu-adsk: We are checking for engine not being ```None```in the [maya engine](https://github.com/shotgunsoftware/tk-maya/pull/7/files#r34182112) but not here. Is this a bug? Another argument in favor of refactoring this logic in a base class.

python/tk_nuke/menu_generation.py
```diff
@@ -44,19 +42,17 @@ def create_menu(self):
...
-        nuke_menu = nuke.menu("Nuke")
-        menu_handle = nuke_menu.addMenu(self._menu_name) 
+        # create main Shotgun menu
+        menu_handle = nuke.menu("Nuke").addMenu(self._menu_name)
         
-        # the right click menu that is displayed when clicking on a pane 
-        self._pane_menu = nuke.menu("Pane") 
         # create tank side menu
-        node_menu_handle = nuke.menu("Nodes").addMenu(self._menu_name, icon=self.tank_logo)
+        node_menu_handle = nuke.menu("Nodes").addMenu(self._menu_name, icon=self._shotgun_logo)
```
jfboismenu-adsk: Shouldn't the shotgun logo resource be something provided by core and ideally setup by the engine?

jfboismenu-adsk: I know the menu logic is not centralized, but one first nice step would be to centralize the icon resources at the minimum.

python/tk_nuke/menu_generation.py
```diff
@@ -138,17 +132,16 @@ def destroy_menu(self):
...
         # will crash). clearMenu() seems to work on both v6 and v7.
+        menus = ["Nuke", "Pane", "Nodes"]
+
+        for menu in menus:
+            # find the menu and iterate over all items
+            for mh in nuke.menu(menu).items():
+                # look for the shotgun menu
+                if mh.name() == self._menu_name:
+                    # and clear it
+                    mh.clearMenu()
```
jfboismenu-adsk: Missing a break it seems again.

python/tk_nuke/panels.py
```diff
@@ -0,0 +1,228 @@
...
+                             should not take any parameters.
+        """
+        
+        # first, store the widget class on the sgtk object
+        # and key it by id. This is because we then pass the class
+        # name as a string into Nuke, and we need a way to uniquely refer
+        # back to the class object.
+        #
+        # Once this attribute is set, it means that you can access the
+        # class from sgtk.panel_id_name
```
jfboismenu-adsk: comment seems out of sync with the actual variables being set.

python/tk_nuke/panels.py
```diff
@@ -0,0 +1,228 @@
...
+            bundle.log_debug("Reparented existing toolkit widget.")
+        
+        # Add the widget to our current layout
+        self.layout.addWidget(self.toolkit_widget)
+        bundle.log_debug("Added toolkit widget to panel hierarchy")
+        
+        # now, the close widget logic does not propagate correctly
+        # down to the child widgets. When someone closes a tab or pane,
+        # QStackedWidget::removeWidget is being called, which merely takes
+        # our widget out of the layout and hides it. So it will stay resindent
```
jfboismenu-adsk: resident

python/tk_nuke/panels.py
```diff
@@ -0,0 +1,228 @@
...
+        # tab itself and have that call our widget so that we can close ourselves.
+        # note that we search for the tab widget by unique id rather than going
+        # up in the widget hierarchy, because the hiearchy has not been properly
+        # established at this point yet. 
+        for widget in QtGui.QApplication.allWidgets():
+            if widget.objectName() == panel_id:
+                filter = CloseEventFilter(widget)
+                filter.parent_closed.connect(self._on_parent_closed)
+                widget.installEventFilter(filter)
+                bundle.log_debug("Installed close-event filter watcher on tab %s" % widget)
```
jfboismenu-adsk: break :)

python/tk_nuke/panels.py
```diff
@@ -0,0 +1,228 @@
...
+                filter = CloseEventFilter(widget)
+                filter.parent_closed.connect(self._on_parent_closed)
+                widget.installEventFilter(filter)
+                bundle.log_debug("Installed close-event filter watcher on tab %s" % widget)     
+        
+    def _find_panel_tab(self, widget):
+        """
+        Helper method.
+        Given a tk panel widget, traverse upwards in the hierarchy and
+        attempt to locate the tab widget.
```
jfboismenu-adsk: A comment explaining why the tab widget might be missing would be appreciated.

engine.py
manneohrstrom: Ok, so the new implementation now centralizes this aspect. It's an indirect change after some deliberation - i guess the consequence is that this now goes away both in maya and nuke.

engine.py
manneohrstrom: I am not totally sure I totally understand your comment here. Nevertheless, this nuke specific panel restore callback is now gone in the reshuffled panel structure.

python/tk_nuke/menu_generation.py
manneohrstrom: Some stuff is centralized - for example the standard banner on top of the UI dialogs. In the case of engines, logos and resources are often pretty specific and selected specifically to suit the needs of a particular engine. I think this is the reason we haven't tried to standardize things. I think it's better to wait with standardizing any icons until we do a big menu overhaul. That way we'll have a better, clearer picture of all the different moving pieces we need to take into account for that work.

python/tk_nuke/panels.py
manneohrstrom: fixed!

python/tk_nuke/panels.py
manneohrstrom: Added a comment explaining :)

engine.py
```diff
@@ -197,6 +217,84 @@ def log_error(self, msg):
         
     
     ##########################################################################################
+    # panel interfaces
+
+    def show_panel(self, panel_id, title, bundle, widget_class, *args, **kwargs):
+        """
+        Shows a panel in a way suitable for this engine. The engine will attempt to
+        integrate it as seamlessly as possible into the host application. If the engine does
```
jfboismenu-adsk: That second sentence is superfluous, we know that the Nuke engine does support panels.

engine.py
```diff
@@ -197,6 +217,84 @@ def log_error(self, msg):
...
+        import tk_nuke
+        
+        # now look for the tank._callback_from_non_pane_menu property
+        # if this exists, that is an indication that this call comes from 
+        # a non-pane menu, like the Shotgun menu. 
+        #
+        # In this case, we need to explicitly specify where to host the window
+        #
+        try:
+            non_pane_callback = tank._callback_from_non_pane_menu
```
jfboismenu-adsk: Wouldn't initializing this to False be simpler instead of raising an AttributeError on the first read? Is there anything to be gained from this?

python/tk_nuke/panels.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        Constructor
+        """
+        
+        # first, call the base class and let it do its thing.
+        QtGui.QWidget.__init__(self)
+        
+        # pick up members from the global sgtk namespace
+        # where it was set by the NukePanelWidget code above
+        panel_id = sgtk._current_panel_id
```
jfboismenu-adsk: I've seen this pattern in other places in Toolkit, but I never got to ask why that pattern is needed. A comment explaining why would be nice.

jfboismenu-adsk: I mean, couldn't have these be set as "static" class members instead?

manneohrstrom: I changed the code to use a class method. It's a little bit more code, but it's a bit clearer what is going on! :)

engine.py
manneohrstrom: I tweaked this to check for the presence of this flag rather than it's value. This way, the menu system can clean up after itself rather than permanently polluting the tank module namespace :)

engine.py
manneohrstrom: Made the comment more nuke specific :)

engine.py
```diff
@@ -197,6 +217,87 @@ def log_error(self, msg):
...
+        Additional parameters specified will be passed through to the widget_class constructor.
+        """
+        # note! not using the import as this confuses nuke's calback system
+        # (several of the key scene callbacks are in the main init file...)
+        import tk_nuke
+        
+        # create the panel
+        panel_widget = tk_nuke.NukePanelWidget(bundle, title, panel_id, widget_class, *args, **kwargs)
+        
+        if hasattr(tank, "_callback_from_non_pane_menu"):
```
jfboismenu-adsk: Could this have been solved by having a "static" on the class instead, just like you did for the UI widget earlier?

manneohrstrom: I did look into it, and I don't think so. There is no easy access to this object from the menu callbacks. I also have to refocus at this point so is it okay for this to stay the way it is. If you have strong objections to the current approach, please let me know and I will look into it more, otherwise there are bigger :fish: to fry.

PR #12, opened 2015-10-13
30562 nuke hiero merge

thebeeland:
Merges the Nuke and Hiero engines.


engine.py
```diff
@@ -23,200 +19,240 @@
...
+    """
+    An engine that supports Nuke 6.3v5+ and Hiero 9.0v*+.
+    """
+    # Define the different areas where menu events can occur in Hiero.
+    (HIERO_BIN_AREA, HIERO_SPREADSHEET_AREA, HIERO_TIMELINE_AREA) = range(3)
 
-    ##########################################################################################
-    # init
+    def __init__(self, *args, **kwargs):
+        self._hiero = nuke.env.get("hiero")
```
manneohrstrom: is this a boolean? In that case I think it should be saying `self._hiero_enabled` or something like that. `self._hiero` sounds to me like it contains the hiero "object".

engine.py
```diff
@@ -23,200 +19,240 @@
...
+
+    @property
+    def has_ui(self):
+        """
+        Whether Nuke is running a GUI session.
+        """
+        return self._ui_enabled
+
+    @property
+    def hiero(self):
```
manneohrstrom: This feels ambiguous to me. Feels like it would be clearer if it said `hiero_enabled` or `hiero_active` or something along those lines to emphasize that it returns a bool.

engine.py
```diff
@@ -23,200 +19,240 @@
...
-            if (self._ui_enabled 
-                and not "TANK_NUKE_ENGINE_INIT_NAME" in os.environ
-                and nuke_version[0] >= self.get_setting("compatibility_dialog_min_version", 10)):
-                nuke.message("Warning - Shotgun Pipeline Toolkit!\n\n%s" % msg)
-                           
-            # and log the warning
-            self.log_warning(msg)
-            
-        # now prepare tank so that it will be picked up by any new processes
+        # Now prepare tank so that it will be picked up by any new processes
```
manneohrstrom: We shouldn't take this bit out. That's something we use to tell people what max version we support! Instead, bump the numbers to the current version we are testing against, so that all future versions will report the warning.

engine.py
```diff
@@ -23,200 +19,240 @@
...
-        # we also need to pass the path to the python folder down to the init script
-        # because nuke python does not have a __file__ attribute for that file
+        # We also need to pass the path to the python folder down to the init script
+        # because nuke python does not have a __file__ attribute for that file.
         local_python_path = os.path.abspath(os.path.join( os.path.dirname(__file__), "python"))
         os.environ["TANK_NUKE_ENGINE_MOD_PATH"] = local_python_path
-        
-    
+
+    @property
```
manneohrstrom: out of curiosity, why move the properties up here? I don't mind, but just to let you the previous setup was intentional i think and the idea was to group all the init stuff at the top and after that put properties.

engine.py
```diff
@@ -23,200 +19,240 @@
...
+        if self.has_ui:
+            # Note! not using the import as this confuses nuke's calback system
             # (several of the key scene callbacks are in the main init file...)            
             import tk_nuke
 
             menu_name = "Shotgun"
             if self.get_setting("use_sgtk_as_menu_name", False):
                 menu_name = "Sgtk"
-            
-            # create menu
```
manneohrstrom: why remove this comment? maybe we can put it back?

engine.py
```diff
@@ -23,200 +19,240 @@
...
-            # a global list of registered panels, try to locate
-            # the one it needs and then run the callback         
-            for (panel_id, panel_dict) in self.panels.iteritems():
-                nukescripts.panels.registerPanel(panel_id, panel_dict["callback"])
-            
-        # iterate over all apps, if there is a gizmo folder, add it to nuke path
+
+            if self.hiero:
+                import hiero
+                def set_project_root(event):
```
manneohrstrom: i think we should avoid defining methods like this if we can - sometimes we do, but it's more if it's a tiny little helper that is only of local relevance - almost like a lambda. This doesn't feel like that - seems like it should be a proper part of the engine?

Also, maybe there should be two methods `init_hiero` and `init_nuke` that this method calls?. That way maybe we could avoid the return inside the hiero if clause too?

engine.py
```diff
@@ -23,200 +19,240 @@
...
+                                )
+                            )
+                            p.setProjectRoot(self.tank.project_path)
+                hiero.core.events.registerInterest(
+                    'kAfterNewProjectCreated',
+                    set_project_root,
+                )
+                # We don't need to get into the Gizmo stuff after this block
+                # since that is specific to Nuke.
+                return
```
manneohrstrom: i think we should avoid this if possible - it easily missed and may cause confusion in the future if we make further changes.

engine.py
```diff
@@ -23,200 +19,240 @@
...
         """
-        Return the QWidget parent for all dialogs created through show_dialog & show_modal.
+        Return the QWidget parent for all dialogs created through
+        show_dialog and show_modal.
         """
-        # see https://github.com/shotgunsoftware/tk-nuke/commit/35ca540d152cc5357dc7e347b5efc728a3a89f4a 
-        # for more info. There have been instability issues with nuke 7 causing various crashes, so 
-        # window parenting on Nuke versions above 6 is currently disabled.
+        # See https://github.com/shotgunsoftware/tk-nuke/commit/35ca540d152cc5357dc7e347b5efc728a3a89f4a 
+        # for more info. There have been instability issues with nuke 7 causing
```
manneohrstrom: is this a PEP8 change? Just curious.

engine.py
```diff
@@ -23,200 +19,240 @@
...
+        # various crashes, so window parenting on Nuke versions above 6 is
+        # currently disabled.
         if nuke.env.get("NukeVersionMajor") > 6:
             return None
         return super(NukeEngine, self)._get_dialog_parent()
 
-    ##########################################################################################
```
manneohrstrom: Do we need to remove this? I think it is meant to make the code more readable, but maybe you disagree?

engine.py
```diff
@@ -23,200 +19,240 @@
...
-    # logging interfaces
-    
     def log_debug(self, msg):
         if self.get_setting("debug_logging", False):
-            # print it in the console
```
manneohrstrom: why remove this comment? I think it adds value.

engine.py
```diff
@@ -23,200 +19,240 @@
...
             msg = "Shotgun Debug: %s" % msg
+            if self.hiero:
+                import hiero
+                hiero.core.log.setLogLevel(hiero.core.log.kDebug)
+                hiero.core.log.debug("Shotgun: %s" % msg)
```
manneohrstrom: I think this will print out `Shotgun: Shotgun Debug: xxx`

engine.py
```diff
@@ -23,200 +19,240 @@
...
+            # 
+            # In order to have these emitted, we would need to turn them on by doing a 
+            # hiero.core.log.setLogLevel(hiero.core.log.kInfo)
+            #
+            # However this has call been omitted on purpose because too much output 
+            # from hiero to stdout/stderr is causing problems with the browser plugin
+            # causing it to hang or crash. By keeping the info and warning logging off
+            # we are avoiding such hangs and working around this known issue.
+            import hiero
+            hiero.core.log.info("Shotgun: %s" % msg)
```
manneohrstrom: same comment here about the double `shotgun` being printed.

engine.py
```diff
@@ -23,200 +19,240 @@
...
         print msg
-        
+
     def log_warning(self, msg):
         msg = "Shotgun Warning: %s" % msg
-        # print it in the nuke error console
```
manneohrstrom: i don't think we should remove any of these comments.

engine.py
```diff
@@ -23,200 +19,240 @@
...
+                self.__setup_favorite_dirs()
+                
+                # Register all panels with nuke's callback system
+                # this will be used at nuke startup in order
+                # for nuke to be able to restore panels 
+                # automatically. For all panels that exist as
+                # part of saved layouts, nuke will look through
+                # a global list of registered panels, try to locate
+                # the one it needs and then run the callback.
+                for (panel_id, panel_dict) in self.panels.iteritems():
```
manneohrstrom: Looks like panel functionality is only available for nuke mode. If this is intentional, i think there needs to be a comment explaining why it's nuke only.

engine.py
```diff
@@ -269,34 +312,26 @@ def show_panel(self, panel_id, title, bundle, widget_class, *args, **kwargs):
...
-        
-        # return the created widget
-        return panel_widget     
-        
-    
-    ##########################################################################################
-    # managing favorite dirs            
+        return panel_widget                
     
     def __setup_favorite_dirs(self):
```
manneohrstrom: so there seems to be a bunch of separate nuke and hiero methods and logic. Depending on the direction we are going with the final engine, wondering if it's worth splitting this up in two separate python files, one for nuke and one for hiero, for simplicity? But maybe that's overkill? Or maybe that's counter productive given the direction we are headed with a fully merged engine?

info.yml
```diff
@@ -13,6 +13,62 @@
 # expected fields in the configuration file for this engine
 configuration:
 
+    bin_context_menu:
```
manneohrstrom: These settings are all missing a `default_value`

python/tk_nuke/menu_generation.py
```diff
@@ -21,291 +18,665 @@
...
 class MenuGenerator(object):
     """
-    Menu generation functionality for Nuke
+    The factory base class for Nuke based menu generators. Since Nuke
+    now provides multiple modes (Nuke, Nuke Studio, and Hiero), we are
+    required to supply menu generation logic for each. Instantiating
+    this class directly will provide the correct implementation for the
+    currently running Nuke mode.
     """
+    def __new__(cls, *args, **kwargs):
```
manneohrstrom: we are doing `if self.hiero` in many places in the engine already. Why not just have this as an (yet another) if statement in the engine instead? That seems clearer to me.

python/tk_nuke/menu_generation.py
```diff
@@ -21,291 +18,665 @@
...
+                    # skip favourites since they are already on the menu
+                    cmd_obj.add_command_to_menu(self._menu_handle)
+
+# -----------------------------------------------------------------------------
+
+class NukeMenuGenerator(MenuGenerator):
+    """
+    A Nuke specific menu generator.
+    """
+    def __init__(self, *args, **kwargs):
```
manneohrstrom: What's the benefit of having `*args, **kwargs` here?

python/tk_nuke/menu_generation.py
```diff
@@ -21,291 +18,665 @@
...
+            os.path.join(
+                engine_root_dir,
+                "resources",
+                "sg_logo_blue_32px.png",
+            ),
+        )
+
+    def create_menu(self):
+        """
+        Creates the "Shotgun" menu in Nuke.
```
manneohrstrom: Could the code for creating the shotgun menu be shared between the hiero and nuke modes?

manneohrstrom:
Cool stuff! Had some questions about things!

One more question:

the hiero engine has a `get_menu_selection()` method that is heavily used by apps (hiero hooks) to figure out where a menu click was coming from (which context menu) - see for example: https://github.com/shotgunsoftware/tk-multi-workfiles/blob/master/hooks/scene_operation_tk-hiero.py#L106 I suspect we need this method in the nuke engine as well if we want the apps to behave the same way!

Awesome stuff! Cheers!

engine.py
thebeeland: Goof on my part. It shouldn't have been removed. It's going back in!

engine.py
thebeeland: Just me flailing around and not paying attention to ordering. I'll clean that up and also add back in the banner comments to help divide up the different "sections" of the class better.

engine.py
thebeeland: Agreed. It came over from the Hiero engine, but I've made it a staticmethod on the class now (once I push). It'll be the same end result but without the local function definition.

engine.py
thebeeland: I've split the Hiero- and Nuke-specific routines for the various init methods out and call those from the "hero" init methods. That makes it MUCH easier to tell what's going on. Good suggestion. :)

engine.py
thebeeland: I'm a grammar nazi!

engine.py
thebeeland: I'm adding them back in now that I've also properly grouped together related methods/properties. The class is now easier to read!

engine.py
thebeeland: I should have replaced the comments in the log methods with better comments instead of just nuking (hahaha pun!) what was there. I've done so now!

engine.py
thebeeland: I more like the earlier idea of having nuke/hiero methods and calling those from the generic "hero" methods. In a number of cases there will be shared code across all "modes" of Nuke, so having completely separate classes doesn't feel much better than having separate engines entirely. It would still mean making code changes in multiple places were we to add features/enhacements/bug fixes that apply across the board.

python/tk_nuke/menu_generation.py
thebeeland: Yeah, I think that's fine. This was the first thing that I merged and I originally had grand plans of making it all transparent and not needing if/else statements everywhere, but in the end I went with the simpler approach with the rest of the engine. I'll clean this up and make it more straightforward.

python/tk_nuke/menu_generation.py
thebeeland: It makes little difference in this case, so I'm now passing the args explicitly. The advantage is that since the Nuke/Hiero specific __init__ methods don't require any arguments themselves, you can pass whatever is received up to the base class. If the __init__ args of the base class change in the future then there's no need to update its subclasses because those new args are passed up the chain automatically via args/kwargs.

In this case it's overkill since it's a very simple, self-contained class structure.

thebeeland:
Thanks for catching get_menu_selection(). Pretty big oversight on my part!

engine.py
thebeeland: Definitely agreed! I'm also going to rename has_ui to be ui_enabled to match. In the near future we will then also have a studio_enabled property to go along with them.

python/tk_nuke/menu_generation.py
thebeeland: Unfortunately it looks as though the Hiero mode of Nuke still makes use of the previous Hiero-specific method of building/managing menus and menu items. Attempting to use the "nuke" handle doesn't work in Hiero mode. As such, the implementations have to stay separate.

thebeeland: I should note, however, that the Nuke menu creation logic will work just fine when we get to Nuke Studio, so we shouldn't need a third implementation there.

engine.py
manneohrstrom: Wait - `has_ui` is part of the engine base class and is used all over the place. That needs to remain the same (see https://support.shotgunsoftware.com/entries/95441127#has_ui%20property abd https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/engine.py#L383 )

We also need to be backwards compatible with these kinds of changes :)

manneohrstrom:
Thanks for addressing the comments so quickly! Doing second pass now!

engine.py
```diff
@@ -23,200 +19,285 @@
...
+        self._ui_enabled = nuke.env.get("gui")
+        super(NukeEngine, self).__init__(*args, **kwargs)
+
+    #####################################################################################
+    # Properties
 
-    ##########################################################################################
-    # init
+    @property
+    def ui_enabled(self):
```
manneohrstrom: i replied to one of the earlier comments - but probably good to add it here just so it doesn't get "folded away" by github: The `has_ui` is part of the official engine interface so we cannot easily change that method! It's a shame beacuse `ui_enabled` works nicely with `hiero_enabled` (and `has_hiero` sounds really weird to me!). But we probably need to keep `has_ui` for consistency and backwards compatibility....

engine.py
```diff
@@ -23,200 +19,285 @@
...
-            # and the warning dialog isn't overriden by the config:
+            # Show nuke message if in UI mode, this is the first time the engine has been started
+            # and the warning dialog isn't overriden by the config.
             if (self._ui_enabled 
                 and not "TANK_NUKE_ENGINE_INIT_NAME" in os.environ
                 and nuke_version[0] >= self.get_setting("compatibility_dialog_min_version", 10)):
                 nuke.message("Warning - Shotgun Pipeline Toolkit!\n\n%s" % msg)
                            
-            # and log the warning
+            # Log the warning.
```
manneohrstrom: I agree that this comment may not be needed :-)

engine.py
```diff
@@ -23,200 +19,285 @@
...
-            
-        # now prepare tank so that it will be picked up by any new processes
+
+        # Make sure we are not running Nuke PLE!
+        if nuke.env.get("ple"):
+            self.log_error("The Nuke Engine does not work with the Nuke PLE!")
+            return
+
+        # Do our mode-specific initializations.
+        if self.hiero_enabled:
```
manneohrstrom: nice!

engine.py
```diff
@@ -23,200 +19,285 @@
...
+        """
+        self._last_clicked_selection = []
+        self._last_clicked_area = None
+
+    def init_engine_nuke(self):
+        """
+        The Nuke-specific portion of engine initialization.
+        """
+        # Now check that there is a location on disk which corresponds to the context.
+        if self.context.project is None:
```
manneohrstrom: I think we can move this up into `init_engine` - i think it's a valid check for both engines, we just didn't do it previously for hiero for some reason. I guess you could argue that the check is a bit overkill, i don't think this error message fires very often given how toolkit is set up...

engine.py
```diff
@@ -269,48 +357,106 @@ def show_panel(self, panel_id, title, bundle, widget_class, *args, **kwargs):
...
+        """
+        return self._last_clicked_area
+
+    #####################################################################################
+    # General Utilities
+
+    @staticmethod
+    def set_project_root(event):
+        """Ensure any new projects get the project root or default startup 
+        projects get the project root set
```
manneohrstrom: i think it would be useful to document the event parameter here in the doc string. It's not totally obvious what sort of thing gets passed in via that parameter.

manneohrstrom:
CR looking awesome. Nice work Sir! Good go to - just minor comments and one bigger about `has_ui`.

PR #13, opened 2015-12-14
Adds bootstrapping logic for Nuke, Nuke Studio, and Hiero.

thebeeland:


python/startup/Python/Startup/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+# 
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
```
robblau: Spacing throughout looks odd in this file.  Make sure we're doing 4 spaces per indent level.

python/startup/Python/Startup/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+def _setup_sgtk(output_handle):
+	"""
+	Extracts the necessary information from the environment and starts
+	the tk-nuke engine.
+	"""
+	try:
+		import tank
+	except Exception, e:
+		output_handle("Shotgun: Could not import sgtk! Disabling: %s" % e)
```
robblau: With a generic exception catch like this, it is safer to end with ```% str(e)``` or ```% (e,)``` to make sure that the exception class isn't also a tuple (leading to an error where not all arguments are converted).  This same issue happens a few times below as well.

PR #14, opened 2016-01-08
Avoid destroy_menu to raise when a Menu is already empty

JeanChristopheMorin:
```python
[13:05.57] ERROR: Shotgun Error: Could not restart the engine!

'MenuItem' object has no attribute 'clearMenu'
The current environment is shot_step.

Code Traceback:
  File ".../studio/install/core/python/tank/platform/__init__.py", line 109, in restart
    engine.destroy()
  File ".../studio/install/core/python/tank/platform/engine.py", line 435, in destroy
    self.destroy_engine()
  File ".../studio/install/engines/app_store/tk-nuke/v0.4.1/engine.py", line 246, in destroy_engine
    self._menu_generator.destroy_menu()
  File ".../studio/install/engines/app_store/tk-nuke/v0.4.1/python/tk_nuke/menu_generation.py", line 396, in destroy_menu
    mh.clearMenu()
```

To avoid this error when a menu (Nuke, Pane or Node) is already empty.

### DIFF
--- a/python/tk_nuke/menu_generation.py
+++ b/python/tk_nuke/menu_generation.py
@@ -391,7 +391,7 @@ class NukeMenuGenerator(BaseMenuGenerator):
             # Find the menu and iterate over all items.
             for mh in nuke.menu(menu).items():
                 # Look for the shotgun menu.
-                if mh.name() == self._menu_name:
+                if isinstance(mh, nuke.Menu) and mh.name() == self._menu_name:
                     # Clear it.
                     mh.clearMenu()
 


kporangehat:
@JeanChristopheMorin,  we've ticketed this up internally and will pull this in shortly. Do you have a specific way to reproduce the issue we can use for testing?

Thanks for the PR Jean-Christophe.

cheers,
kp

JeanChristopheMorin:
This was an error cause by some of our code which was playing with Shotgun's Node menu. There was a double cleanup of the menu, our and yours. Our cleanup was done before yours and it was causing this error. But I'd say let's keep on going with this, since it's only a safety fix. We never know what the user will do!

nebukadhezer:
@kporangehat 
Any idea when this will be deployed ?
I am having this issue and was close to submit the same pr as @JeanChristopheMorin ...
Though I am not playing with menus or anything, I am just having an empty menu in the toolbar from sg... 
Seems as Nuke is "downgrading" a menu to an menuItem but I cannot reproduce why...
So @JeanChristopheMorin 's fix seems like a good solution to me.

thanks
Johannes

thebeeland:
I'm looking at getting this into master now. At a glance, your approach looks fine to me. We've recently seen the same problem crop up, so I'm not having any trouble reproducing the problem.

PR #19, opened 2016-02-03
fix menu generation if self._menu_name is a MenuItem

ylepaih:
I have some issue with destroy_menu
sometimes "Shotgun" is a MenuItem, not a Menu and clearMenu() call FAILED

Remove MenuItem named self._menu_name before clearMenu prevent this crash
Best Regards

### DIFF
--- a/python/tk_nuke/menu_generation.py
+++ b/python/tk_nuke/menu_generation.py
@@ -559,6 +559,10 @@ class NukeMenuGenerator(BaseMenuGenerator):
         menus = ["Nuke", "Pane", "Nodes"]
         for menu in menus:
             # Find the menu and iterate over all items.
+            
+            # remove self.menu_name if exist as MenuItem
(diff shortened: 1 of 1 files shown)


ylepaih:
I saw, this is a duplicate pull request with : https://github.com/shotgunsoftware/tk-nuke/pull/14

When the Shotgun Menu is Empty, it appears as MenuItem and crash because MenuItem have no method clearMenu

The problem is the same, but the patch is little bit different : I propose to try to remove Shotgun MenuItem

kporangehat:
@ylepaih,

Thanks for the PR. We discussed this and decided we're going to go with the solution in #14 just because it's slightly closer to our current code. Just to double check though, will that solution cover the issue you're seeing?

Thanks Yann. Really appreciate the contributions!

ylepaih:
Hi KP
The pull request  #14 will fix my problem
Thanks
Yann

PR #20, opened 2016-02-03
Enhance __tank_startup_node_callback context modification

ylepaih:
Default behaviour in __tank_startup_node_callback is a little annoying

- If you load a file out of the pipe, Shotgun menu appears disabled and you can't open a pipeline file (instead of using file>open dialog which is not pipeline compliant)

- Another point, I found dangerous to inherited context from the previous session when you click on file>open. This behaviour could generate save with unwanted context.

To fix these two problems, I propose to switch on "project" context as default behaviour for file>new/clear and unsupported file
It enables open/inject dirty scene in the pipeline and never lose the shotgun menu

Best Regards
Yann

### DIFF
--- a/python/tk_nuke/__init__.py
+++ b/python/tk_nuke/__init__.py
@@ -152,19 +152,13 @@ def __tank_startup_node_callback():
             return
             
         if nuke.root().name() == "Root":
-            # file->new
-            # base it on the context we 'inherited' from the prev session
-            # get the context from the previous session - this is helpful if user does file->new
+            # file->new or clear
+            # load a default project context
             project_root = os.environ.get("TANK_NUKE_ENGINE_INIT_PROJECT_ROOT")
             tk = tank.Tank(project_root)
-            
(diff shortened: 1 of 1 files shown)


kporangehat:
Agreed this is definitely something we want to handle better. In general this would be a decent approach but we want to be sure that we don't change the default behavior magically without the user knowing. This was the initial reason for disabling the menus and engine. If you change the context without also having a way to notify the user, then you're in a dangerous situation again. 

Yes, there's the same dangerous situation you mention when using `FIle > Open` but we need to think through this on a deeper level to come up with a really graceful way of handling this case while also keeping the user informed as to what's happening.

Thank you for submitting this. It's sparked up some good discussion internally as well, and as I mentioned, we do plan to handle this better. In the meantime, you're welcome to fork the engine and use it internally if you like. Just keep in mind, you'll need to maintain it for updates etc.

I'm going to close this out for now, but feel free to share any other thoughts you may have. Thanks again.

PR #25, opened 2016-07-27
#35878: Fixes boostrap regression.

thebeeland:
The engine's module path is now added to sys.path on launch. Also removes old startup directory that's no longer in use.


manneohrstrom:
The `menu.py` file says

```
"""		
This file is loaded automatically by Nuke when file->new or file->open is triggered.		

Note however that this does not automatically run *the first* time nuke starts up,		
despite being an auto-start script. 		
		
This is to overcome the fact that file->new in nuke typically results in a completely new		
process being launched.		
```

How does file->new work now? Maybe we could add some comments to `sgtk_startup.py` (or somewhere else if that makes more sense) with an overview of the entire bootstrap and context switching process? It's weird and intricate in Nuke and I am not sure it's really very well documented anywhere - comments are scattered all over the place! :)

thebeeland:
Ugh...I didn't notice the comment about file->new resulting in a new process. I'll have to put the menu.py back in place.

I'll open an internal ticket to add documentation of the entire bootstrap process for all three Nuke/NS/Hiero. It's definitely worth doing.

PR #29, opened 2017-02-14
#39873 Updated Nuke version to support 10.5

mathurf:
Did update the Nuke version to support 10.5
Did also update the minimum version supported since we are no more
supporting Nuke 6

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -163,7 +163,7 @@ class NukeEngine(tank.platform.Engine):
             self.log_error(msg)
             return
 
-        # Versions > 9.0 have not yet been tested so show a message to that effect.
+        # Versions > 10.5 have not yet been tested so show a message to that effect.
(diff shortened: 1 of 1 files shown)


engine.py
```diff
@@ -155,19 +155,16 @@ def init_engine(self):
...
             return
-        elif nuke_version[0] == 6 and nuke_version[1] < 3:
-            self.log_error(msg)
-            return
-        elif nuke_version[0] == 6 and nuke_version[1] == 3 and nuke_version[2] < 5:
+        elif nuke_version[0] == 7 and nuke_version[1] == 0 and nuke_version[2] < 10:
             self.log_error(msg)
             return
 
         # Versions > 9.0 have not yet been tested so show a message to that effect.
```
thebeeland: [minor] Comment is no longer correct.

PR #31, opened 2017-03-01
For #40775, scanning for Nuke-based products

jfboismenu:
This implements the SoftwareLauncher interface in order to launch Nuke on every platform. It also introduces a unit test suite that allows to ensure the launcher is providing the correct flags and icons. Note that at this point the nuke engine is not launched once Nuke is running.


startup.py
```diff
@@ -0,0 +1,383 @@
...
+def _template_to_product_name(template):
+    """
+    Converts a display name template name to a product name.
+
+    For example, "NukeX %s Non-commercial" becomes "NukeX Non-commercial".
+    """
+    return template.replace("%s", "").replace("  ", " ").strip()
+
+
+def _format(template, tokens):
```
jfboismenu: I've done this to remain Python 2.5 compliant. Should I have bother with the zero config stuff?

startup.py
```diff
@@ -0,0 +1,373 @@
+# Copyright (c) 2017 Shotgun Software Inc.
```
josh-t: I don't think the icon above is `256x256` pixels. :P

startup.py
josh-t: My vote is **NO**.

startup.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+
+    # Named regex strings to insert into the executable template paths when
+    # matching against supplied versions and variants. Similar to the glob
+    # strings, these allow us to alter the regex matching for any of the
+    # variable components of the path in one place
+
+    COMPONENT_REGEX_LOOKUP = {
+        "version": r"(?P<version>[\d.v]+)",
+        "variant": r"(?P<variant>[\w\s]+)",
```
josh-t: should we change these to `product`?

startup.py
```diff
@@ -0,0 +1,373 @@
...
+                self.logger.debug(
+                    "Found %s matches: %s" % (
+                        len(matching_paths),
+                        matching_paths
+                    )
+                )
+
+        return executable_matches
+
+    def _find_products(self):
```
josh-t: suggest `_find_software_versions`. at least until we have a better name for `SoftwareVersion` object.

startup.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        For each software executable that was found, get the software products for it.
+
+        :returns: Generator that will iterate on each SoftwareVersion that was found.
+        """
+        executable_matches = self._find_executables(self.EXECUTABLE_MATCH_TEMPLATES[sys.platform])
```
josh-t: My original implementation broke this up into 2 `for` loops, but it's not necessary. See the houdini/photoshop implementations if you are interested in reducing to that form.

startup.py
```diff
@@ -0,0 +1,373 @@
...
+
+            # build the glob pattern by formatting the template for globbing
+            glob_pattern = _format(match_template, dict((key, "*") for key in self.COMPONENT_REGEX_LOOKUP))
+            self.logger.debug(
+                "Globbing for executable matching: %s ..." % (glob_pattern,)
+            )
+            matching_paths = glob.glob(glob_pattern)
+            if matching_paths:
+                # found matches, remember this association (template: matches)
+                executable_matches[match_template] = [x.replace("\\", "/") for x in matching_paths]
```
josh-t: maybe add some comments here about the replacement?

startup.py
```diff
@@ -0,0 +1,373 @@
...
+                self.logger.debug("Processing path: %s" % (executable_path,))
+
+                match = executable_regex.match(executable_path)
+
+                if not match:
+                    self.logger.debug("Path did not match regex.")
+                    continue
+
+                for product in self._extract_products_from_path(executable_path, match.groupdict()):
+                    yield product
```
josh-t: generator. I like it.

startup.py
jfboismenu: Ah, you're right. Seems like the macOS review app doesn't export the right resolution when trying to save the icon. I had to click on the right icon and drag it out from the app and into the folder to get the right resolution.

#badUX

startup.py
```diff
@@ -0,0 +1,253 @@
...
+    engine with the current context in the new session of Nuke.
+    """
+
+    # Named regex strings to insert into the executable template paths when
+    # matching against supplied versions and products. Similar to the glob
+    # strings, these allow us to alter the regex matching for any of the
+    # variable components of the path in one place
+
+    COMPONENT_REGEX_LOOKUP = {
+        "version": r"(?P<version>[\d.v]+)",
```
josh-t: Should we simplify to not use the `?P<>` syntax now that we can disguise it via the helper method?

PR #33, opened 2017-03-14
For #40775, nuke plugin

jfboismenu:
Implements the nuke plugin.

- Implements a Nuke plugin that can be used with the software launcher as well as standalone.
- implements 0.18 based logging.
- implements the automatic_context_switch setting on the engine, even tough not technically needed for zero config since it uses an entirely different code path and way to restart the engine on File->New|Open, i.e. it does nothing and simply reloads the plugin.
- Moves most of the Toolkit Classic bootstrap code from `python/startup` to `classic_startup` so that it sits at the same level as the `plugins` code.
- `python/startup/boostrap.py`needs to stay there because that is being invoked by the legacu multi-launchapp
- `startup/menu.py` has been moved to `classic_startup/restart/menu.py` to encapsulate its use.


classic_startup/menu.py
```diff
@@ -0,0 +1,20 @@
...
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
```
jfboismenu: This is not a new file but a file moved from `python/startup/menu.py` to this location.

engine.py
```diff
@@ -223,30 +233,19 @@ def init_engine_nuke(self):
...
-        # Add our startup path to the nuke init path
-        startup_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "startup"))
-        tank.util.append_path_to_env_var("NUKE_PATH", startup_path)        
-    
-        # We also need to pass the path to the python folder down to the init script
-        # because nuke python does not have a __file__ attribute for that file.
-        local_python_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "python"))
-        os.environ["TANK_NUKE_ENGINE_MOD_PATH"] = local_python_path
 
-    def pre_app_init(self):
```
jfboismenu: `pre_app_init`'s job was to register the nuke file->New|Open callbacks for Nuke, but it turns out that post `start_engine` call we are already doing that so wait until we get there to invoke it so the call only happens once.

engine.py
```diff
@@ -334,6 +333,7 @@ def post_app_init_hiero(self, menu_name="Shotgun"):
             # (several of the key scene callbacks are in the main init file).
             import tk_nuke
             import hiero
+            from hiero.core import env as hiero_env
```
jfboismenu: Full code of the orignal method if [here](https://github.com/shotgunsoftware/tk-nuke/blob/master/engine.py#L334-L364) This import was missing, which turned out to throw an error everytime we tried to log usage stats, which itself is exception safe so it silently failed for bad reasons all the time.

engine.py
```diff
@@ -462,57 +462,30 @@ def post_context_change(self, old_context, new_context):
     #####################################################################################
     # Logging
 
-    def log_debug(self, msg):
```
jfboismenu: 0.18, oh yeah!

python/tk_nuke/menu_generation.py
```diff
@@ -366,7 +366,7 @@ def _add_context_menu(self):
         action = ctx_menu.addAction("Jump to Shotgun")
         action.triggered.connect(self._jump_to_sg)
 
-        if self.context.filesystem_locations:
+        if ctx.filesystem_locations:
```
jfboismenu: This was a mistake in the original Jump To Filesystem implementation I submitted today. Should have tested in Hiero first!

startup.py
```diff
@@ -241,4 +241,64 @@ def prepare_launch(self, exec_path, args, file_to_open=None):
...
+                required_env["TANK_CONTEXT"] = sgtk.Context.serialize(self.context)
+                required_env["TANK_ENGINE"] = self.engine_name
+
+            self.logger.debug("Launch environment: %s", pprint.pformat(required_env))
+            self.logger.debug("Launch arguments: %s", required_args)
+
+        return LaunchInformation(exec_path, required_args, required_env)
+
+    @contextlib.contextmanager
+    def temporary_import(self, module_path, module_name):
```
jfboismenu: Useful pattern to import a module temporarily. Most engines probably need this. Maybe this needs to be refactored?

startup.py
```diff
@@ -241,4 +241,64 @@ def prepare_launch(self, exec_path, args, file_to_open=None):
...
+
+        # Get the basic environment variables to launch Nuke. Do this by reusing Toolkit classic's
+        # bootstrap code.
+        startup_python_path = os.path.join(
+            self.disk_location,
+            "python",
+            "startup"
+        )
+
+        with self.temporary_import(startup_python_path, "bootstrap") as bootstrap:
```
manneohrstrom: not sure i understand this part - why do we need to import `python/startup`? Is it because we keep stuff in that file for backwards compatibility with older launch apps that might call it? In that case, could we do it the other way, and have the old legacy bootstrap interface in `python/startup` call `startup.py` (the new official and cleaned up interface) instead of the other way around? This feels like a very complicated approach - would love to discuss if there is a simpler way forward.

engine.py
manneohrstrom: does that mean hiero usage analytics are skewed?

engine.py
manneohrstrom: YES!!!

python/startup/bootstrap.py
```diff
@@ -43,5 +60,68 @@ def bootstrap(engine_name, context, app_path, app_args, extra_args):
...
+        os.environ.get("TANK_FILE_TO_OPEN")
+    )
+
+    for name, value in env_vars.iteritems():
+        sgtk.util.append_path_to_env_var(name, value)
+
     return (app_path, app_args)
 
+
+def get_plugin_startup_env(plugin_names, app_path, app_args, file_to_open):
```
manneohrstrom: seems better to keep in `startup.py`?

engine.py
jfboismenu: Probably, yes. I'll ping the rest of the team about it.

startup.py
jfboismenu: I'll try to see if we can do it the other way around, this is a very good idea.

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
```diff
@@ -0,0 +1,179 @@
...
+    #   module which we can use to retrieve the location of core and add it
+    #   to the pythonpath.
+    # - If we are running toolkit as part of a larger zero config workflow
+    #   and not from a standalone workflow, we are running the plugin code
+    #   directly from the engine folder without a bundle cache and with this
+    #   configuration, core already exists in the pythonpath.
+    # If the engine is currently being restarted, restore
+
+    # now see if we are running stand alone or in situ
+    try:
```
jfboismenu: This is pretty much the same boiler plate as in the Houdini engine.

engine.py
```diff
@@ -223,30 +235,20 @@ def init_engine_nuke(self):
...
-    def pre_app_init(self):
-        """
-        Called at startup, but after QT has been initialized.
-        """
-        if self.hiero_enabled or self.studio_enabled:
-            return
+        # If we're in Toolkit classic mode, we need to backup a few things in order to be able
+        # to restart the engine after a File->New|Open
+        if not self.in_plugin_mode:
+            os.environ["TANK_NUKE_ENGINE_INIT_PROJECT_ROOT"] = self.tank.project_path
```
thebeeland: Are these new env vars? If so, should they be "SGTK" instead of "TANK"?

plugins/basic/menu.py
```diff
@@ -0,0 +1,39 @@
...
+        plugin_bootstrap.bootstrap(plugin_root_path)
+    except Exception, e:
+        import traceback
+        stack_trace = traceback.format_exc()
+
+        message = "Shotgun Toolkit Error: %s" % (e,)
+        details = "Error stack trace:\n\n%s" % (stack_trace)
+
+        import nuke
+        nuke.error(message)
```
thebeeland: Are we guaranteed to be in Nuke (not Nuke Studio or Hiero) when this is called? We'll let a NotImplementedError bubble up, otherwise.

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
```diff
@@ -0,0 +1,179 @@
...
+    # - If we are running the plugin built as a stand-alone unit,
+    #   try to retrieve the path to sgtk core and add that to the pythonpath.
+    #   When the plugin has been built, there is a sgtk_plugin_basic_nuke
+    #   module which we can use to retrieve the location of core and add it
+    #   to the pythonpath.
+    # - If we are running toolkit as part of a larger zero config workflow
+    #   and not from a standalone workflow, we are running the plugin code
+    #   directly from the engine folder without a bundle cache and with this
+    #   configuration, core already exists in the pythonpath.
+    # If the engine is currently being restarted, restore
```
thebeeland: Incomplete thought here?

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
```diff
@@ -0,0 +1,179 @@
...
+    ``tk-nuke`` and ``tk-nukestudio`` respectively.
+    """
+    engine_name = os.environ.get("SHOTGUN_ENGINE")
+    if not engine_name:
+        # FIXME: We are assuming that the configuration inside the standalone plugin
+        # has an engine instance with these name. These values should probably come
+        # from the plugin's info.yml file.
+        try:
+            import hiero # noqa
+        except:
```
thebeeland: Any reason why this isn't `except ImportError:` ?

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
```diff
@@ -0,0 +1,179 @@
...
+            engine_name = "tk-nuke"
+        else:
+            engine_name = "tk-nukestudio"
```
thebeeland: Mostly just a question out of curiosity. Are we specifically not supporting Hiero, or are we just referencing a tk-nukestudio engine instance in the config for both NS and Hiero?

startup.py
```diff
@@ -67,7 +66,7 @@ class NukeLauncher(SoftwareLauncher):
         ],
         "win32": [
             # C:/Program Files/Nuke10.0v5/Nuke10.0.exe
-            "C:/Program Files/Nuke{version}/Nuke{major_minor_version}.exe",
+            "C:\\Program Files\\Nuke{version}\\Nuke{major_minor_version}.exe",
```
thebeeland: Use a raw string here instead of escaping the backslashes:
```
r"C:\Program Files\Nuke..."
```

tests/test_startup.py
```diff
@@ -22,7 +22,7 @@
 
 
 repo_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
-print "tk-nuke repoistory root found at %s." % repo_root
+print "tk-nuke repository root found at %s." % repo_root
 
 
 class TestStartup(TankTestBase):
```
thebeeland: TESTS!!!!

engine.py
```diff
@@ -181,15 +189,19 @@ def init_engine(self):
...
+            return
+        elif nuke.env.get("nc"):
+            self.log_error("The Nuke Engine does not work with Nuke Non-Commercial!")
             return
 
-        # Now check that there is a location on disk which corresponds to the context.
-        if self.context.project is None:
+        # Now check that we are at least in a project context. Note that plugin mode
+        # does not require this check since it can operate at the site level.
+        if not self.in_plugin_mode and self.context.project is None:
```
manneohrstrom: could we lift the restriction for the engine completely here rather than just for plugins?

engine.py
```diff
@@ -223,30 +235,20 @@ def init_engine_nuke(self):
...
-        """
-        if self.hiero_enabled or self.studio_enabled:
-            return
+        # If we're in Toolkit classic mode, we need to backup a few things in order to be able
+        # to restart the engine after a File->New|Open
+        if not self.in_plugin_mode:
+            os.environ["TANK_NUKE_ENGINE_INIT_PROJECT_ROOT"] = self.tank.project_path
+
+            # Add our startup path to the nuke init path
+            startup_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "classic_startup", "restart"))
```
manneohrstrom: `self.disk_location` makes the code more readable IMHO

info.yml
```diff
@@ -13,6 +13,12 @@
 # expected fields in the configuration file for this engine
 configuration:
 
+    automatic_context_switch:
```
manneohrstrom: no list of plugins setting in nuke?

plugins/basic/menu.py
```diff
@@ -0,0 +1,39 @@
+import inspect
```
manneohrstrom: missing std header

plugins/basic/menu.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import sys
+
```
manneohrstrom: would be great with a docstring here explaining what this file is and when it is being executed.

plugins/basic/menu.py
```diff
@@ -0,0 +1,39 @@
...
+
+def plugin_startup():
+    """
+    Initializes the Toolkit plugin for Nuke.
+    """
+
+    current_file_path = os.path.abspath(
```
manneohrstrom: i think deserves a comment :)

plugins/basic/menu.py
```diff
@@ -0,0 +1,39 @@
...
+    plugin_root_path = os.path.dirname(current_file_path)
+
+    # the plugin python path will be just below the root level. add it to
+    # sys.path
+    plugin_python_path = os.path.join(plugin_root_path, "python")
+    sys.path.insert(0, plugin_python_path)
+
+    # now that the path is there, we can import the plugin bootstrap logic
+    try:
+        from tk_nuke_basic import plugin_bootstrap
```
manneohrstrom: i think this will prevent certain tk reloads to work. might be worth poining out in comments

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
manneohrstrom: no hiero support in zero conf as of now

python/startup/bootstrap.py
```diff
@@ -1,47 +1,66 @@
...
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
```
manneohrstrom: we should add a comment here explaining what this entry point is and who is calling it.

Am i right thinking that this is now just a legacy pathway for the old tk-multi-launchapp? 
We should add some notes to say that this will be deprecated at some point in the future when noone is using the older launchapp.

python/startup/bootstrap.py
```diff
@@ -1,47 +1,66 @@
...
+
+logger = sgtk.LogManager.get_logger(__name__)
+
+
+def _get_current_module_path():
+    """
+    Returns the current module's absolute path.
+    """
+    return os.path.dirname(
+        os.path.abspath(sys.modules[_get_current_module_path.__module__].__file__)
```
manneohrstrom: why doesn't just `__file__` work here? Maybe worth a comment

python/startup/bootstrap.py
```diff
@@ -1,47 +1,66 @@
...
+            _get_current_module_path(), # tk-nuke/python/startup
+            "..", # tk-nuke/python
+            ".." # tk-nuke
+        )
+    )
+
 
 def bootstrap(engine_name, context, app_path, app_args, extra_args):
     """
+    Invoked by the non-SoftwareLauncher based startup code.
```
manneohrstrom: ah yes, this comment needs to be more prominent i think! and we should mention the fact that it's legacy! so that someone coming into the code looking at this file thinking "what is this and when is it being used/being run" can see that easily!

python/startup/bootstrap.py
```diff
@@ -1,47 +1,66 @@
...
     Prepares for the bootstrapping process that will run during startup of
     Nuke, Hiero, and Nuke Studio.
 
     .. NOTE:: For detailed documentation of the bootstrap process for Nuke,
               Hiero, and Nuke Studio, see the engine documentation in
               `tk-nuke/engine.py`.
     """
-    import tank
+    startup_file = os.path.join(_get_bundle_root(), "startup.py")
+    module = imp.load_source(uuid.uuid4().hex, startup_file)
```
manneohrstrom: i normally add little comments to highlight whenever i do something slightly special or non-standard

python/tk_nuke/__init__.py
```diff
@@ -190,9 +190,15 @@ def tank_ensure_callbacks_registered():
...
-        nuke.addOnScriptSave(__tank_on_save_callback)
-        g_tank_callbacks_registered = True
 
+    import sgtk
+    engine = sgtk.platform.current_engine()
+
+    # Register only if we're missing an engine (to allow going from disabled to something else)
+    # or if the engine specifically requests for it.
+    if not engine or engine.get_setting("automatic_context_switch"):
+        global g_tank_callbacks_registered
```
manneohrstrom: maybe add some logging here as well?

startup.py
```diff
@@ -241,4 +240,118 @@ def prepare_launch(self, exec_path, args, file_to_open=None):
...
+
+        self.logger.debug("Launch environment: %s", pprint.pformat(required_env))
+        self.logger.debug("Launch arguments: %s", required_args)
+
+        return LaunchInformation(exec_path, required_args, required_env)
+
+    # Do not remove or rename this method. It is being called by older versions
+    # of the tk-multi-launchapp from tk-nuke/python/startup/bootstrap.py.
+    # Also, because this method is being invoke from the bootstrap.py file, it doesn't
+    # have access to the disk_location property and as such needs to be passed in.
```
manneohrstrom: great comment!

tests/test_startup.py
jfboismenu: Not just tests: [CI](https://github.com/shotgunsoftware/tk-nuke/tree/develop/zero_config#documentation)!

engine.py
jfboismenu: They are not. They look new because the diff is bad. These are for classic mode.

plugins/basic/menu.py
jfboismenu: We are not supporting Hiero at the moment with TaaP, so we can always import Nuke.

You just made me realize I hadn't tested with Nuke Studio however and it didn't work. I'll push an update. It was minimal.

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
jfboismenu: Old comment, should have been discarded. Thanks!

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
jfboismenu: Good point. I've removed the code however because there is no such thing an a nuke studio section in the config at the moment.

engine.py
jfboismenu: Forgot about that, thanks!

engine.py
jfboismenu: Not sure I follow. How can an engine in Toolkit Classic run without a project? That doesn't seem feasible at the moment. I can always remove the code as I don't think it was reachable in the first place.

info.yml
jfboismenu: Oops! Added.

plugins/basic/menu.py
jfboismenu: You caught me copy pasting code from Houdini, which is not required for Nuke/Nuke Studio. I've cleaned this up.

python/startup/bootstrap.py
jfboismenu: I don't know honestly. That complicated code was already there before. Let me check.

plugins/basic/menu.py
```diff
@@ -8,11 +23,11 @@ def plugin_startup():
     Initializes the Toolkit plugin for Nuke.
     """
 
-    current_file_path = os.path.abspath(
```
jfboismenu: This was copy pasted from Houdini, but not required for Nuke.

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
```diff
@@ -156,14 +155,8 @@ def __get_engine_name():
...
         # FIXME: We are assuming that the configuration inside the standalone plugin
-        # has an engine instance with these name. These values should probably come
-        # from the plugin's info.yml file.
-        try:
-            import hiero # noqa
-        except:
-            engine_name = "tk-nuke"
-        else:
-            engine_name = "tk-nukestudio"
+        # has an engine instance with this name.
```
jfboismenu: There is only one section in the config at the moment, so the engine to bootstrap is always tk-nuke.

PR #34, opened 2017-03-16
For #40775, nuke plugin polish 

jfboismenu:
This is mostly a polish pass for the Nuke plugin. It adds asynchronous bootstrap and runs the actions found in the `run_at_startup` setting.

There's a lot of issues with Nuke Studio 9. I ended up having to disable both the asynchronous bootstrap and run_at_startup feature otherwise I would get crashes at startup. :(

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -410,7 +410,9 @@ class NukeEngine(tank.platform.Engine):
                 # (for example if you do file->open or file->new)
                 tank.util.append_path_to_env_var("NUKE_PATH", app_gizmo_folder)
 
-        self._run_commands_at_startup()
+        # Nuke Studio 9 really doesn't like us starting things at startup, do don't.
+        if not (nuke.env.get("NukeVersionMajor") == 9 and nuke.env.get("studio")):
+            self._run_commands_at_startup()
 
         try:
             self.log_user_attribute_metric("Nuke version", nuke.env.get("NukeVersionString"))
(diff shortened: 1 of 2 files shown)


engine.py
```diff
@@ -410,12 +410,75 @@ def post_app_init_nuke(self, menu_name="Shotgun"):
...
+        if not (nuke.env.get("NukeVersionMajor") == 9 and nuke.env.get("studio")):
+            self._run_commands_at_startup()
+
         try:
             self.log_user_attribute_metric("Nuke version", nuke.env.get("NukeVersionString"))
         except:
             # ignore all errors. ex: using a core that doesn't support metrics
             pass
 
+    def _run_commands_at_startup(self):
```
jfboismenu: All this code until the FIXME is a big copy paste from Maya. Or what it Houdini? (which copied it from Maya in any case). We should create a ticket about refactoring this bit.

engine.py
```diff
@@ -410,12 +410,75 @@ def post_app_init_nuke(self, menu_name="Shotgun"):
                 # (for example if you do file->open or file->new)
                 tank.util.append_path_to_env_var("NUKE_PATH", app_gizmo_folder)
 
+        # Nuke Studio 9 really doesn't like us starting things at startup, do don't.
```
josh-t: typo: `do` => `so`

josh-t: suggest: 

```
# Nuke Studio 9 really doesn't like us running commands at startup, so don't.
```

plugins/basic/python/tk_nuke_basic/plugin_bootstrap.py
```diff
@@ -101,6 +104,10 @@ def __launch_sgtk(base_config, plugin_id, bundle_cache):
     sgtk_logger = sgtk.LogManager.get_logger("plugin")
     sgtk_logger.debug("Booting up toolkit plugin.")
 
+    sgtk_logger.debug("Executable: %s", sys.executable)
+    sgtk_logger.debug("Studio?: %s", nuke.env.get("studio"))
```
josh-t: Suggest something more descriptive for debugging:

```
sgtk_logger.debug("Studio environment variable set? %s", nuke.env.get("studio"))
sgtk_logger.debug("Hiero environment variable set? %s", nuke.env.get("hiero"))

```

PR #36, opened 2017-03-30
41525 Remove legacy log_* methods and use engine.logger.* instead

gomezll:
Remove legacy log_* methods and use engine.logger.* instead

These are changes for task: http://internal.shotgunstudio.int/page/10287#Ticket_41525_%5Btk-nuke%5D%3A%20Finish%20switching%20to%20new%20style%20logging%20in%20the%20engine.

This is my first code review, please let me know if there's anyone else i should add as a reviewer.


gomezll:
At DWA we used to prefer having more than 1 reviewer. But one of the reviewers was designated the 'primary', so that person would be in charge of making sure the changes were reviewed thoroughly and of guiding along the pull request through the approval process. The role was sort of an arbiter of the reviewers, making sure issues were addressed to reasonable satisfaction and that the process was completed successfully before handing final approval.

gomezll:
Pushed up your suggested tweaks @jfboismenu . Thanks for review

PR #38, opened 2017-06-22
#43239 Fix to preserve original NUKE_PATH

pscadding:
Should now keep any original paths stored in the NUKE_PATH env var, when adding the shotgun toolkit specific paths. It will now join the shotgun toolkit path along with other paths instead of overwriting them.

This fixes a bug that appears to be a regression.


pscadding:
I've not done much building of tests before. I had to modify the tests to get them to complete successfully. Hopefully, my approach was correct?

tests/test_startup.py
```diff
@@ -245,26 +245,39 @@ def _mock_folder_listing(self):
         else:
             yield
 
-    def _get_plugin_environment(self, dcc_path):
+    def _join_existing_environment(self, dcc_path, startup_path):
+        """
+        Returns the pre-existing environment paths for dcc_path, os.pathsep joined with the startup_paths
+        """
```
manneohrstrom: please add params to the docstring

tests/test_startup.py
```diff
@@ -245,26 +245,39 @@ def _mock_folder_listing(self):
         else:
             yield
 
-    def _get_plugin_environment(self, dcc_path):
+    def _join_existing_environment(self, dcc_path, startup_path):
+        """
+        Returns the pre-existing environment paths for dcc_path, os.pathsep joined with the startup_paths
```
manneohrstrom: nit: startup path rather than startup paths?

startup.py
```diff
@@ -342,6 +342,10 @@ def _compute_environment(cls, app_path, app_args, startup_paths, file_to_open):
         elif "nukestudio" in app_path.lower() or "--studio" in app_args:
             env["HIERO_PLUGIN_PATH"] = os.pathsep.join(startup_paths)
         else:
+            existing_paths = os.environ.get("NUKE_PATH")
```
manneohrstrom: it can be confusing to repurpose variables like you have done here, e..g first `startup_paths` is the list of paths to plugins that need to be added, then it suddenly becomes that PLUS any existing paths. this can be easy to miss for a future code maintainer who is trying to also make sure of the `startup_paths` parameter.

Something along these lines is clearer - also note added comments which help tell the story a bit:

```
# get any existing nuke path to custom gizmos, scripts etc.
existing_nuke_path_str = os.environ.get("NUKE_PATH") or ""
existing_nuke_path_list = existing_nuke_path_str.split(os.pathsep)

# prepend any toolkit extensions in order to ensure the right integrations execute
modified_nuke_path_list = startup_paths + existing_nuke_path_list

env["NUKE_PATH"] = os.pathsep.join(modified_nuke_path_list)
```

tests/test_startup.py
```diff
@@ -245,26 +245,39 @@ def _mock_folder_listing(self):
...
+        """
+        startup_path = [startup_path]
+        existing_paths = os.environ.get(dcc_path)
+        # combine the existing dcc_path paths, with the sgtk startup_paths only if there are preexisting paths
+        startup_paths = existing_paths.split(os.pathsep) + startup_path if existing_paths else startup_path
+        # now combine the list of paths with the os specific separator
+        return os.pathsep.join(startup_paths)
+
+    def _get_plugin_environment(self, dcc_path, ):
```
manneohrstrom: if possible, seems we should add a test where the we already have a NUKE_PATH set to test the order in which the new NUKE_PATH is being assembled.

tests/test_startup.py
pscadding: Yeah I'm not sure. I put startup path as all the functions that would call this would pass a single string path like `"/my_path/"`

I then put it in a list in the method. I wasn't sure if it should expect a list of strings  or a pathseparated string (in which case it would be startup_paths), or just a single path in a string.
I've based it upon the latter, so I went with startup_path.

tests/test_startup.py
```diff
@@ -299,75 +314,106 @@ def test_nuke_studio(self):
...
     def test_nukex(self):
         """
         Ensures NukeX LaunchInformation is correct.
         """
-        for engine_instance, is_classic in self._get_engine_configurations():
-            self._test_launch_information(
-                engine_instance, "NukeX.app", "", "/file/to/open",
-                self._get_nuke_environment(is_classic=is_classic)
-            )
+        # loop over an empty path and a test path to simulate having no pr-existing environment path
```
manneohrstrom: i think we should leave the launch tests as is and add a new, separate test that tests the `NUKE_PATH` stuff separately. we are basically testing the same startup logic for hiero, nuke, nukex etc. Instead i suggest just create a `test_nuke_path_override()` (or something like that) test which specifically tests that logic. 

In a full-on TDD world, this is how you'd typically approach it - find the bug, first write a new failing tests which reproduces the problem, and then add implementation which makes the test pass.

tests/test_startup.py
jfboismenu: Instead of modifying the test so that it can handle when the environment variable is set prior to running the test maybe you could clear the environment variable from os.environ during setUp. Simply override [setUp](https://github.com/shotgunsoftware/tk-core/blob/master/tests/python/tank_test/tank_test_base.py#L193) from the base class (but don't forget to call it!)

tests/test_environment_paths.py
```diff
@@ -0,0 +1,197 @@
...
+        Checks against pre-existing and non-existent scenarios.
+        :return: None
+        """
+        nuke_path_1 = os.path.join(tempfile.gettempdir(), "gizmo_1")
+        nuke_path_2 = os.path.join(tempfile.gettempdir(), "gizmo_1")
+        plugin_path = os.path.join(repo_root, "plugins", "basic")
+
+        try:
+            # test override
+            os.environ["HIERO_PLUGIN_PATH"] = os.pathsep.join([nuke_path_1, nuke_path_2])
```
jfboismenu: Using [tank_test.temp_env_var](https://github.com/shotgunsoftware/tk-core/blob/master/tests/python/tank_test/tank_test_base.py#L103) instead to set and reset env var.

tests/test_environment_paths.py
```diff
@@ -0,0 +1,197 @@
...
+        patch = mock.patch.dict("os.environ", {"TK_NUKE_REPO_ROOT": repo_root})
+        self.addCleanup(patch.stop)
+        patch.start()
+
+        # Setup the fixture. This will take the configuration at fixtures/config inside this
+        # repo.
+        self.setup_fixtures()
+
+        # Update the mocked hierarchy to contain the user folder on Linux.
+        if sys.platform == "linux2":
```
jfboismenu: Not needed for your tests. There is no mocked hierarchy.

PR #40, opened 2017-08-05
For #43349: Highjacks the launch to force an engine instance for Nuke Studio.

thebeeland:
This resolves the problem of us providing a single Software entity for the Nuke family of products, but needing different configuration for Nuke Studio as compared to Nuke.

This is considered a medium-term fix, as Rob and I have discussed a Product->engine_instance map likely controlled via launchapp hook. That will require more work than can be done right this moment, so this is the fix for the time being.


thebeeland:
The long-term solution is ticket is 44069 on internal. I'll not be merging this PR right now, as we intend to have the better solution in place before releasing the default2 config to clients.

PR #45, opened 2017-10-23
SPLAT-83: Added methods to retrieve script path and dependencies.

rupalkhilari:
Added implementation for the following engine methods:
* get_session_path() - Returns the absolute file path of the nuke script if saved, else returns an empty string.
* get_session_dependencies() - Returns the file path dependencies of the current nuke script for the global file frame range. 

Linked with: https://github.com/shotgunsoftware/tk-core/pull/544

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -925,6 +925,7 @@ class NukeEngine(tank.platform.Engine):
         :returns: Path to the current session if it has been saved to disk,
                   else returns None.
         """
+
         if self.hiero_enabled or self.studio_enabled:
             import hiero
 
@@ -937,9 +938,8 @@ class NukeEngine(tank.platform.Engine):
             # only one selection.
             if len(selection) != 1:
                 hiero.core.log.error("Cannot determine the current session " +
-                                     "dependencies as the selected project " +
-                                     "could not be deduced from 0 or multiple " +
-                                     "selections.")
+                                     "filename as a single project selection " +
+                                     "is required to do so.")
                 return None
 
             # if the user has selected a non-project element,
@@ -956,9 +956,8 @@ class NukeEngine(tank.platform.Engine):
                 # apparently bins can be without projects (child bins)
                 hiero.core.log.error("Cannot determine the current session " +
                                      "as the project could not be deduced.")
-                return None
-
-            session_path = project.path()
+            else:
+                session_path = project.path()
         else:
             session_path = nuke.root().knob("name").value()
 
@@ -1000,45 +999,40 @@ class NukeEngine(tank.platform.Engine):
             # only one selection.
             if len(selection) != 1:
                 hiero.core.log.error("Cannot determine the current session " +
-                                     "dependencies as the selected project " +
-                                     "could not be deduced from 0 or multiple " +
-                                     "selections.")
-                return []
-
-            # if the user has selected a non-project element,
-            # return None
-            if not isinstance(selection[0], hiero.core.Bin):
+                                     "dependencies as a single project selection " +
+                                     "is required to do so.")
+            # if the user has selected a non-project element, return an
+            # empty list
+            elif not isinstance(selection[0], hiero.core.Bin):
                 hiero.core.log.error("Cannot determine the current session " +
                                      "dependencies because an element other " +
                                      "than a project is selected.")
-                return []
+            else:
+                project = selection[0].project()
 
-            project = selection[0].project()
-
-            if project is None:
-                # apparently bins can be without projects (child bins)
-                hiero.core.log.error("Cannot determine the current session " +
-                                     "dependencies as the project could not " +
-                                     "be deduced from the selection.")
-                return []
-
-            # collect dependencies for the selected project.
-            for clip in project.clips():
-                mediaSource = clip.mediaSource()
-                for info in mediaSource.fileinfos():
-                    # replace forward slashes with the OS-
-                    # specific path separator to make
-                    # Nuke happy on Windows.
-                    dependencies.append(
-                        {
-                            "path": info.filename().replace(
-                                "/",
-                                os.path.sep
-                            ),
-                            "engine": self.name,
-                            "type": "MediaSource",
-                        }
-                    )
+                if project is None:
+                    # apparently bins can be without projects (child bins)
+                    hiero.core.log.error("Cannot determine the current session " +
+                                         "dependencies as the project could not " +
+                                         "be deduced from the selection.")
+                else:
+                    # collect dependencies for the selected project.
+                    for clip in project.clips():
+                        mediaSource = clip.mediaSource()
+                        for info in mediaSource.fileinfos():
+                            # replace forward slashes with the OS-
+                            # specific path separator to make
+                            # Nuke happy on Windows.
+                            dependencies.append(
+                                {
+                                    "path": info.filename().replace(
+                                        "/",
+                                        os.path.sep
+                                    ),
+                                    "engine": self.name,
+                                    "type": "MediaSource",
+                                }
+                            )
         else:
             # create a list of Read-like nodes in the session
             read_nodes = [node for node in nuke.allNodes()
--- a/engine.py
+++ b/engine.py
@@ -927,9 +927,7 @@ class NukeEngine(tank.platform.Engine):
         # retrieve the name of the nuke file. In case of a saved file, the
         # absolute path of the file is returned, for an unsaved file,
         # an empty string is returned.
-        path = nuke.root().knob("name").value()
-
-        return path
+        return nuke.root().knob("name").value()
 
     def get_session_dependencies(self):
         """
@@ -945,8 +943,13 @@ class NukeEngine(tank.platform.Engine):
         start_frame = int(nuke.root().knob("first_frame").value())
         end_frame = int(nuke.root().knob("last_frame").value())
 
-        # retrieve the enabled Write nodes
-        enabled_write_nodes = self._get_enabled_write_nodes()
+        # create a list of enabled write node in the session
+        enabled_write_nodes = []
+        write_nodes = [node for node in nuke.allNodes()
+                       if node.Class() == "Write"]
+        if write_nodes:
+            enabled_write_nodes = [node for node in write_nodes
+                                   if not node.knob("disable").value()]
 
         # find file dependencies for each write node
         for write_node in enabled_write_nodes:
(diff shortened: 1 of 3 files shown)


engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
+
+    def get_session_dependencies(self):
+        """
+         Returns the list of file or folder dependencies for the current
+         session.
+
+         :return: A list of file or folder paths that the current
+                  session depends on.
+         """
+        dependencies = []
```
jfboismenu: I think we should have an engine hook that allows to filter the list of session dependencies so clients can either block certain dependencies from being returned or add ones that we may have missed.

@manneohrstrom @robblau ?

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
+
+    def get_session_path(self):
+        """
+        Returns the path to the current file if it resides on
+        disk. If unsaved, it returns None.
+
+        :return: Path to the current scene if file is saved, else returns
+                 and empty string.
+        """
+        # retrieve the name of the nuke file. In case of a saved file, the
```
jfboismenu: Is that compatible with Nuke Studio and Hiero? I don't know how these environments work. @thebeeland would know more.

rupalkhilari: Looks like it is compatible with Nuke Studio but always returns an empty string in Hiero. The closest I could find is the last loaded project path with `hiero.core.projects()[-1].path()` but there should be something better.

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
                                 type=(nuke.IMAGE | nuke.SCRIPT | nuke.GEO),
                                 icon=icon_path,
                                 tooltip=path)
+
+    #####################################################################################
+    # Script and session related methods
+
+    def get_session_path(self):
```
manneohrstrom: This does not reflect the core interface - a `session=None` parameter is missing!

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
+
+        :return: Path to the current scene if file is saved, else returns
+                 and empty string.
+        """
+        # retrieve the name of the nuke file. In case of a saved file, the
+        # absolute path of the file is returned, for an unsaved file,
+        # an empty string is returned.
+        return nuke.root().knob("name").value()
+
+    def get_session_dependencies(self):
```
manneohrstrom: This does not reflect the core interface - a `session=None` parameter is missing!

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
                                 icon=icon_path,
                                 tooltip=path)
+
+    #####################################################################################
+    # Script and session related methods
+
+    def get_session_path(self):
+        """
+        Returns the path to the current file if it resides on
+        disk. If unsaved, it returns None.
```
manneohrstrom: I think the comment is potentially misleading - it sounds to me like we are saying 'if there are unsaved changed, it returns None'. maybe just to be clear, it could say something like 'if the session has never been saved and isn't associated with a file on disk yet, `None` is returned.

engine.py
manneohrstrom: yes absolutely! We should give a hook the ability to post process the list. This was discussed but my thinking was that we keep it simple for the time being and we can potentially add that in the future. @jfboismenu if you strongly disagree, let's discuss and see if we can add it in now. It's not a big change i think.

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
+
+        :return: Path to the current scene if file is saved, else returns
+                 and empty string.
+        """
+        # retrieve the name of the nuke file. In case of a saved file, the
+        # absolute path of the file is returned, for an unsaved file,
+        # an empty string is returned.
+        return nuke.root().knob("name").value()
```
manneohrstrom: on windows, please ensure that paths with backslashes are returned.

engine.py
manneohrstrom: Please add hiero support before we merge this in!

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
+        """
+         Returns the list of file or folder dependencies for the current
+         session.
+
+         :return: A list of file or folder paths that the current
+                  session depends on.
+         """
+        dependencies = []
+
+        # we shall query for file dependencies for the file frame range
```
manneohrstrom: This seems akin to something like a frustum culling in Maya. I am not sure this is always the behaviour we want. I think the default behaviour should be to return *all* dependencies in the scene, regardless of frame range. @rupalkhilari, did you have a particular use case in mind for the frame range "culling"? I think for a basic, cross-DCC and generic implementation, we want to return all dependencies in the scene. Potentially in the future we can add options to limit the set of returned items. What do you think?

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
+        start_frame = int(nuke.root().knob("first_frame").value())
+        end_frame = int(nuke.root().knob("last_frame").value())
+
+        # create a list of enabled write node in the session
+        enabled_write_nodes = []
+        write_nodes = [node for node in nuke.allNodes()
+                       if node.Class() == "Write"]
+        if write_nodes:
+            enabled_write_nodes = [node for node in write_nodes
```
manneohrstrom: again, i think we want the dependencies to reflect the set of inputs into the scene, regardless of how those are being used. Imagine an application where we want to perform a remote transfer of work from one artist to another - in that case, it probably makes sense to transfer the "disabled" content.

engine.py
```diff
@@ -912,3 +912,58 @@ def __setup_favorite_dirs(self):
...
+            # extract the file path dependencies of all
+            # the Read nodes
+            for node, file_paths in file_dependencies:
+                if node.Class() == "Read":
+                    dependencies.extend(file_paths)
+
+        # filter out the unique paths
+        dependencies = list(set(dependencies))
+
+        return dependencies
```
manneohrstrom: Similar to the maya PR, I think we should return a richer data structure:

```
[
  {"path": "/foo/bar/hello.%04d.jpeg", "engine": "tk-nuke", "type": "Read"},
  {"path": "/foo/bar/hello.%04d.obj", "engine": "tk-nuke", "type": "ReadGeo"},
]
```

engine.py
rupalkhilari: @manneohrstrom I see your point. My only thought for using the minimum set was that the artist shouldn't be waiting on transfer of data that isn't going to be used for the render. The `nuke.Node.fileDependencies()` also requires a start and end frame. But I see your point - it is possible that frame range could be changed, and at that point it might become tricky then.

manneohrstrom: maybe we could just grab all write and writegeo nodes out of the scene? what does `nuke.Node.fileDependencies()` do?

rupalkhilari: Oh, did you mean Read and ReadGeo nodes ? The nuke.Node.fileDependencies() returns all file dependencies from the node graph traversing upwards from that Node. (more info [here](https://docs.thefoundry.co.uk/nuke/110/pythonreference/nuke.Node-class.html#fileDependencies) for a given frame range. That probably makes sense for rendering, but probably not so good for someone to be working on the file.

manneohrstrom: sorry, yes i meant read and readgeo!

rupalkhilari:
@manneohrstrom and @jfboismenu , thanks for the feedback! The comments have been addressed.    Whenever possible could you let me know if there are additional changes to be made ?

engine.py
```diff
@@ -912,3 +912,126 @@ def __setup_favorite_dirs(self):
...
+    #####################################################################################
+    # Script and session related methods
+
+    def get_session_path(self, session=None):
+        """
+        Returns the absolute path to the current file if it resides
+        on disk. If the session has never been saved and isn't
+        associated with a file on disk yet, an empty string is returned.
+
+        :param session: An object representing the active document
```
manneohrstrom: following our discussion with josh, let's drop the session param!

engine.py
```diff
@@ -912,3 +912,126 @@ def __setup_favorite_dirs(self):
...
+
+            # first find what the current project is. Hiero is a multi
+            # project environment so we can ask the engine which
+            # project was last clicked.
+            selection = self.get_menu_selection()
+
+            # to guard against multiple selections, make sure we have
+            # only one selection.
+            if len(selection) != 1:
+                raise tank.TankError("Please select a single project.")
```
manneohrstrom: i think the error message may be confusing for users - perhaps it should just emit a warning instead and say something like Warning: cannot determine the current session filename because multiple projects are currently selected?

engine.py
```diff
@@ -912,3 +912,126 @@ def __setup_favorite_dirs(self):
...
+            if not isinstance(selection[0], hiero.core.Bin):
+                raise tank.TankError("Please select a Hiero project.")
```
manneohrstrom: similar here, i think perhaps just warn and return None

engine.py
```diff
@@ -912,3 +912,126 @@ def __setup_favorite_dirs(self):
...
+
+            session_path = project.path()
+        else:
+            session_path = nuke.root().knob("name").value()
+
+        # replace forward slashes with the OS-specific path separator
+        # to make Nuke happy on Windows.
+        return session_path.replace("/", os.path.sep)
+
+    def get_session_dependencies(self, session=None):
```
manneohrstrom: following our discussion with josh, let's drop the session param!

engine.py
```diff
@@ -912,3 +912,126 @@ def __setup_favorite_dirs(self):
...
+                     "type": "ReadGeo2"
+                     },
+                  ]
+         """
+        dependencies = []
+
+        if self.hiero_enabled or self.studio_enabled:
+            import hiero.core
+
+            # collect dependencies for all projects opened
```
manneohrstrom: i think it's possibly asymmetrical to get all the dependencies here where you only get the file name for one project when you call `get_session_path()` - i think we should apply the same logic as we do above and try to find out what the current session is and then get deps for it.

engine.py
rupalkhilari: Since Hiero does not seem to have the warning log level, I chose `hiero.core.log.error()`, as the info log level is not displayed by default in the console. Do you think `hiero.core.log.info()` would be more appropriate ?

PR #47, opened 2017-11-08
#45353 Secondary publish port for nuke > flame workflow

josh-t:
This PR ports the legacy secondary publish behavior found in `tk-config-flame` to a new publish2 hook in the `tk-nuke` engine. 

The logic can be found in these two files in the flame config:
* https://github.com/shotgunsoftware/tk-config-flame/blob/master/hooks/multi_publish/nuke/secondary_pre_publish.py#L111-L125
* https://github.com/shotgunsoftware/tk-config-flame/blob/master/hooks/multi_publish/nuke/secondary_publish.py#L203-L228

Most of this is copy/paste with some refactoring to fit the new publish plugin API. 

For now, the plugin will only accept if the Shot's clip file exists. The gives us a way to know if Flame is actually being used by the studio. This differs from the legacy behavior where the lack of the Shot clip file would raise an exception. Whether this is the correct behavior for the default2 config is still in discussion, but it doesn't change the functionality. Feel free to weigh in here.

Here's a screenshot of what the plugin looks like in action (Ignore the warning icon). 

image


hooks/tk-multi-publish2/basic/collector.py
```diff
@@ -389,6 +389,10 @@ def collect_sg_writenodes(self, parent_item):
             item.properties["work_template"] = \
                 sg_writenode_app.get_node_render_template(node)
 
+            # store the nuke writenode on the item as well. this can be used by
+            # secondary publish plugins
+            item.properties["nuke_writenode"] = node
```
manneohrstrom: maybe better with `shotgun_writenode` here?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -0,0 +1,618 @@
...
+        if sg_writenode_app:
+            self.logger.debug("Nuke write node app is installed")
+            item.properties["sg_writenode_app"] = sg_writenode_app
+        else:
+            self.logger.debug(
+                "The nuke write node app is not configured for this "
+                "environment. It is required for publishing for Flame."
+                "Plugin %s no accepting item: %s" % (self, item)
+            )
+            accepted = False
```
manneohrstrom: wouldn't it be easier to just return early in all these cases?

hooks/tk-multi-publish2/basic/collector.py
josh-t: Yes, definitely. Will change!

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
josh-t: Yeah, I should change this too.

PR #48, opened 2018-01-30
For #46098: First pass at Flame OpenClip updating when no templates exist.

thebeeland:
There are a number of behavioral changes included, as documented below.

**Input acceptance:**
* If tk-nuke-writenode is in use, only items associated with an instance of that app will be accepted.
* If the writenode app is NOT in use, then all image sequence items will be accepted.

**Clip lookup behavior:**
* If the publish plugin is configured with a "Flame Clip Template" we will attempt to find a clip file matching that template.
* If there is no template configured or the clip file does not exist in the location that the template expects it to, we look for a PublishedFile entity linked to the item's context's entity (most likely a Shot) that matches the PublishedFileType associated with Flame OpenClip publishes.

**Current caveats:**
* We're modifying a PublishedFile's file on disk in place. This is bad for philosophical and technical reasons. The solution we've discussed is creating a new PublishedFile entity at the next available version number, but pointing to the same .clip file on disk. It does not make sense to version the clip file itself, but versioning the PublishedFile entity on top of it will help keep a record of changes, and also better support potential multi-location workflows that rely of PublishedFile entities for data transfer services.
* It appears as though acceptance routines for publish plugins are not re-run when the item's link changes. This is a potential problem in a ZC situation where the user has not changed context in the Panel app prior to launching publish2. In that case, when the acceptance routine is run for the flame clip update plugin, we won't know what Shot to get a clip file from and we'll return a negative. If the user then selects their Shot/Task to publish to from the publish2 UI, we aren't re-run and we're potentially not updating a clip file when we should be. **WORKAROUND:** Set a Shot context using the Panel app prior to launching the Publish app.
* We're taking the first clip PublishedFile from the database and only modifying that one. This means that if multiple .clips are published for a given Shot, only one will be updated. This isn't a regression, as the template-based behavior has a similar caveat, but it might be worth a separate ticket to better support multiple clips if that's a workflow that makes sense for some clients.

screen shot 2018-02-05 at 6 34 37 pm

screen shot 2018-01-29 at 5 40 16 pm

flame_versions_clip

### DIFF
--- a/hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
+++ b/hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
@@ -13,6 +13,8 @@ import os
 import shutil
 import uuid
 import xml.dom.minidom as minidom
+import glob
+import re
 
 import sgtk
 
@@ -274,10 +276,7 @@ class UpdateFlameClipPlugin(HookBaseClass):
         """
 
         # update shot clip xml file with this publish
-        try:
-            self._update_flame_clip(item)
-        except Exception, e:
-            raise("Unable to update Flame clip xml: %s" % (e,))
+        self._update_flame_clip(item)
 
     def finalize(self, settings, item):
         """
@@ -435,82 +434,95 @@ class UpdateFlameClipPlugin(HookBaseClass):
 
         self.logger.info("Updating Flame clip file...")
 
-        # get a handle on the write node app, stored during accept()
-        write_node_app = item.properties["sg_writenode_app"]
-
-        # each publish task is connected to a nuke write node instance. this
-        # value was populated via the collector and verified during accept()
-        write_node = item.properties["sg_writenode"]
-
         # get the clip path as processed during validate()
         flame_clip_path = item.properties["flame_clip_path"]
 
-        # get the fields from the work file
-        render_path = write_node_app.get_node_render_path(write_node)
-        render_template = write_node_app.get_node_render_template(write_node)
-        render_path_fields = render_template.get_fields(render_path)
-        publish_template = write_node_app.get_node_publish_template(write_node)
+        # get a handle on the write node app, stored during accept()
+        write_node_app = item.properties.get("sg_writenode_app")
 
-        # set up the sequence token to be Flame friendly
-        # e.g. mi001_scene_output_v001.[0100-0150].dpx
-        # note - we cannot take the frame ranges from the write node -
-        # because those values indicate the intended frame range rather
-        # than the rendered frame range! In order for Flame to pick up
-        # the media properly, it needs to contain the actual frame data
-
-        # get all paths for all frames and all eyes
-        paths = self.parent.sgtk.paths_from_template(
-            publish_template,
-            render_path_fields,
-            skip_keys=["SEQ", "eye"]
-        )
-
-        # for each of them, extract the frame number. Track the min and the max.
-        # TODO: would be nice to have a convenience method in core for this.
-        min_frame = None
-        max_frame = None
-        for path in paths:
-            fields = publish_template.get_fields(path)
-            frame_number = fields["SEQ"]
-            if min_frame is None or frame_number < min_frame:
-                min_frame = frame_number
-            if max_frame is None or frame_number > max_frame:
-                max_frame = frame_number
-
-        # ensure we have a min/max frame
-        if min_frame is None or max_frame is None:
-            raise Exception(
-                "Couldn't extract min and max frame from the published "
-                "sequence! Will not update Flame clip xml."
+        if not write_node_app:
+            # If we don't have a writenode, we just parse the first frame of the
+            # item's sequence path to get the start and end frames.
+            publish_path_flame = _get_flame_frame_spec_from_path(
+                item.properties["sequence_paths"][0]
             )
 
-        # now when we have the real min/max frame, we can apply a proper
-        # sequence marker for the Flame xml. Note that we cannot use the normal
-        # FORMAT: token in the template system, because the Flame frame format
-        # is not totally "abstract" (e.g. %04d, ####, etc) but contains the
-        # frame ranges themselves.
-        #
-        # the format spec is something like "04"
-        sequence_key = publish_template.keys["SEQ"]
+            if not publish_path_flame:
+                raise Exception(
+                    "Couldn't extract min and max frame from the published "
+                    "sequence! Will not update Flame clip xml."
+                )
+        else:
+            # each publish task is connected to a nuke write node instance. this
+            # value was populated via the collector and verified during accept()
+            write_node = item.properties["sg_writenode"]
 
-        # now compose the format string, eg. [%04d-%04d]
-        format_str = "[%%%sd-%%%sd]" % (
-            sequence_key.format_spec,
-            sequence_key.format_spec
-        )
+            # get the fields from the work file
+            render_path = write_node_app.get_node_render_path(write_node)
+            render_template = write_node_app.get_node_render_template(write_node)
+            render_path_fields = render_template.get_fields(render_path)
+            publish_template = write_node_app.get_node_publish_template(write_node)
 
-        # and lastly plug in the values
-        render_path_fields["SEQ"] = format_str % (min_frame, max_frame)
+            # set up the sequence token to be Flame friendly
+            # e.g. mi001_scene_output_v001.[0100-0150].dpx
+            # note - we cannot take the frame ranges from the write node -
+            # because those values indicate the intended frame range rather
+            # than the rendered frame range! In order for Flame to pick up
+            # the media properly, it needs to contain the actual frame data
 
-        # contruct the final path - because flame doesn't have any windows
-        # support and because the "hub" platform is always linux (with potential
-        # flame assist and flare satellite setups on macosx), request that the
-        # paths are written out on linux form regardless of the operating system
-        # currently running.
-        publish_path_flame = publish_template.apply_fields(
-            render_path_fields,
-            "linux2"
-        )
+            # get all paths for all frames and all eyes
+            paths = self.parent.sgtk.paths_from_template(
+                publish_template,
+                render_path_fields,
+                skip_keys=["SEQ", "eye"]
+            )
+
+            # for each of them, extract the frame number. Track the min and the max.
+            # TODO: would be nice to have a convenience method in core for this.
+            min_frame = None
+            max_frame = None
+            for path in paths:
+                fields = publish_template.get_fields(path)
+                frame_number = fields["SEQ"]
+                if min_frame is None or frame_number < min_frame:
+                    min_frame = frame_number
+                if max_frame is None or frame_number > max_frame:
+                    max_frame = frame_number
+
+            # ensure we have a min/max frame
+            if min_frame is None or max_frame is None:
+                raise Exception(
+                    "Couldn't extract min and max frame from the published "
+                    "sequence! Will not update Flame clip xml."
+                )
+
+            # now when we have the real min/max frame, we can apply a proper
+            # sequence marker for the Flame xml. Note that we cannot use the normal
+            # FORMAT: token in the template system, because the Flame frame format
+            # is not totally "abstract" (e.g. %04d, ####, etc) but contains the
+            # frame ranges themselves.
+            #
+            # the format spec is something like "04"
+            sequence_key = publish_template.keys["SEQ"]
+
+            # now compose the format string, eg. [%04d-%04d]
+            format_str = "[%%%sd-%%%sd]" % (
+                sequence_key.format_spec,
+                sequence_key.format_spec
+            )
+
+            # and lastly plug in the values
+            render_path_fields["SEQ"] = format_str % (min_frame, max_frame)
+
+            # contruct the final path - because flame doesn't have any windows
+            # support and because the "hub" platform is always linux (with potential
+            # flame assist and flare satellite setups on macosx), request that the
+            # paths are written out on linux form regardless of the operating system
+            # currently running.
+            publish_path_flame = publish_template.apply_fields(
+                render_path_fields,
+                "linux2"
+            )
 
         # open up and update our xml file
         xml = minidom.parse(flame_clip_path)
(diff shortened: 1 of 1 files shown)


hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +128,96 @@ def accept(self, settings, item):
...
+            if item.properties.get("sg_writenode"):
+                self.logger.debug("Nuke write node instance found on item.")
+            else:
+                self.logger.debug(
+                    "Unable to identify a nuke writenode instance for this item. "
+                    "Plugin %s no accepting item: %s" % (self, item)
+                )
+                return {"accepted": False}
+
+        # Check to see if we have a template to use. In that case, the
```
josh-t: `then`

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +128,96 @@ def accept(self, settings, item):
...
+        # we'll be trying to find a clip file that that location. If we
```
josh-t: `at that`

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -372,82 +433,96 @@ def _update_flame_clip(self, item):
...
-            sequence_key.format_spec
-        )
+            # for each of them, extract the frame number. Track the min and the max.
+            # TODO: would be nice to have a convenience method in core for this.
+            min_frame = None
+            max_frame = None
+            for path in paths:
+                fields = publish_template.get_fields(path)
+                frame_number = fields["SEQ"]
+                if min_frame is None or frame_number < min_frame:
```
josh-t: You could also just collect all the frame numbers, then call `min(frames)` and `max(frames)` instead of doing the check on each iteration.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -372,82 +433,96 @@ def _update_flame_clip(self, item):
...
+                    min_frame = frame_number
+                if max_frame is None or frame_number > max_frame:
+                    max_frame = frame_number
+
+            # ensure we have a min/max frame
+            if min_frame is None or max_frame is None:
```
josh-t: This would only be the case if there were no frames, right?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -589,6 +740,11 @@ def _generate_flame_clip_name(context, publish_fields):
 
     name = ""
 
+    # If we have template fields passed in, then we'll try to extract
+    # some information from them. If we don't, then we fall back on
+    # some defaults worked out below.
+    publish_fields = publish_fields or dict()
```
thebeeland: @manneohrstrom Any suggestions on how we might better handle the situation here without any template fields to make use of? I'm especially concerned about this always defaulting to version 0 when we aren't using a template. Maybe there's some data I can be pulling that publish2 has already figured out?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -570,6 +645,82 @@ def _update_flame_clip(self, item):
...
+        return None
+
+    # We need to get all files that match the pattern from disk so that we
+    # can determine what the min and max frame number is. We replace the
+    # frame number or token with a * wildcard.
+    glob_path = "%s%s" % (
+        re.sub(match.group(2), "*", root),
+        ext,
+    )
+    files = glob.glob(glob_path)
```
thebeeland: @josh-t I realize now that this might be overkill. Is it safe to use the sequence_paths property from the publish item? Is it always going to include the entire list of frames on disk?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +131,125 @@ def accept(self, settings, item):
...
-            return {"accepted": False}
 
-        # ---- Ensure the clip path exists
+            if clip_publishes:
+                self.logger.debug("Found clip(s): %s" % clip_publishes)
+
+            for clip_publish in clip_publishes:
+                self.logger.debug("Checking existence of OpenClip: %s" % clip_publish)
+                try:
+                    flame_clip_path = resolve_publish_path(publisher.sgtk, clip_publish)
```
thebeeland: @manneohrstrom Once I've found a PublishedFile that we're going to use, I can record that in the publish item for later use when we want to "version up" by creating a new PublishedFile. Do we have some reusable code lying around that will make creating the next available version of a given PublishedFile easier than coding it all out myself in this plugin?

manneohrstrom:
starting to look now!

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +131,125 @@ def accept(self, settings, item):
...
+                except Exception:
+                    self.logger.debug("Unable to resolve path: %s" % clip_publish)
+
+                if flame_clip_path and os.path.exists(flame_clip_path):
+                    self.logger.debug("Found usable OpenClip publish: %s" % flame_clip_path)
+                    break
```
manneohrstrom: i would love if we would iterate over *all* the clips and issue warnings back to the publish loggers to explain that we are not choosing files too. This would be 'grouped' by version number, so say you had 

```
foo_bar.clip v1 /path/to/shot/foo_bar.clip
foo_bar.clip v2 /path/to/shot/foo_bar.clip 
foo_bar.clip v3 /path/to/shot/foo_bar.clip
foo_bar.clip v4 /path/to/shot/foo_bar.clip

baz.clip v1 /path/to/shot/baz.clip
baz.clip v2 /path/to/shot/baz.clip
baz.clip v3 /path/to/shot/baz.clip
```
and we pick foo_bar.clip (v4), we should warn about baz.clip existing but not beng updated. Does that make sense? @josh-t @thebeeland 

Imaing

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
manneohrstrom: Can you just grab it from the `PublishedFile.version_number` and increment it? it shouldn't be a lot of code i would have thought!

```python
# figure out next version number
version_number = 1
if most_recent_publish:
   # there is already a publish in sg
   version_number = most_recent_publish["version_number"] + 1

# create new pub
register_publish(....)
```

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -211,10 +275,7 @@ def publish(self, settings, item):
         """
 
         # update shot clip xml file with this publish
-        try:
-            self._update_flame_clip(item)
-        except Exception, e:
-            raise("Unable to update Flame clip xml: %s" % (e,))
+        self._update_flame_clip(item)
```
manneohrstrom: why was this removed? Will the removal lead to obscure/hard to read error messages in case the xml is malformed?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -372,82 +433,96 @@ def _update_flame_clip(self, item):
...
-                "Couldn't extract min and max frame from the published "
-                "sequence! Will not update Flame clip xml."
+        # get a handle on the write node app, stored during accept()
+        write_node_app = item.properties.get("sg_writenode_app")
+        render_path_fields = None
+
+        if not write_node_app:
+            # If we don't have a writenode, we just parse the first frame of the
+            # item's sequence path to get the start and end frames.
+            publish_path_flame = _get_flame_frame_spec_from_path(
```
manneohrstrom: seems this should be in the path info utils class in publish? What do you think @josh-t ?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
manneohrstrom: when we are not using templates, can't we use the versioning coming from the fact that we are publishing all things in shotgun? So with each change to the clip xml file, we are also creating a published file in sg. I think it makes sense from a workflow perspective if the entries in the clip files reflect the version history in shotgun of the clip file. Does that make sense?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
josh-t: If I'm reading correctly, a method that, given a path with a frame number, returns a path where the frame number is replaced with a spec. We have [this](http://developer.shotgunsoftware.com/tk-multi-publish2/utility.html#tk_multi_publish2.util.get_frame_sequence_path) already, but it's not quite what is needed here. I think we could certainly add something like this, but the more I look at `path_info`, the more I think we need to really consider long term how it fits with the strings/template system. My inclination is to let this live here for the time being.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
josh-t: Currently I don't think so. I took a look and I think this is true for sg writenode-based items and items created via folder drop. But it doesn't look to be the case with regular write nodes. I need to go back and take a closer look and figure out why we're not building that list for those items. That said, you might be able to take advantage of [this](http://developer.shotgunsoftware.com/tk-multi-publish2/utility.html#tk_multi_publish2.util.get_frame_sequences) method here instead of writing your own logic.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
thebeeland: I should add that back. The downside to what was there before is that it smothers the stack, so it's tough to debug the code when something is raised during the XML update. However, the error message is cleaner with the way it was before. I'll set it back up the way it was...I only changed it while debugging.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
thebeeland: This is old code that's just been moved around, but I'll take a quick look.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +131,125 @@ def accept(self, settings, item):
...
-        else:
-            self.logger.debug(
-                "No flame clip template configured. "
-                "Plugin %s no accepting item: %s" % (self, item)
+            flame_clip_fields = item.context.as_template_fields(flame_clip_template)
+            flame_clip_path = flame_clip_template.apply_fields(flame_clip_fields)
+        elif item.context.entity:
+            self.logger.debug("Attempting to find OpenClip publish in %s..." % item.context)
+
+            # TODO: We have a problem. The publish2 plugin acceptance routines
```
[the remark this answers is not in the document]
thebeeland: 👍

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
thebeeland: I'm not going to refactor any of the older logic other than to move it where it needs to be right now.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
thebeeland: At a glance, I think you're correct. However, I'm not going to tinker with any of this old logic other than what I've already had to do.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +131,125 @@ def accept(self, settings, item):
...
+            # In most cases this will not do anything differently than the template
+            # driven solution, because we'll have a single OpenClip published for
+            # the shot, but there might be cases where this heads off a problem.
+            clip_publish_types = ["Flame OpenClip", "Flame Batch OpenClip"]
+            self.logger.debug("Clip publish types to search for: %s" % clip_publish_types)
+            clip_publishes = publisher.shotgun.find(
+                publish_type,
+                [
+                    ["entity", "is", item.context.entity],
+                    ["published_file_type.PublishedFileType.code", "in", clip_publish_types],
```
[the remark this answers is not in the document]
thebeeland: Yep, makes perfect sense.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -209,12 +298,12 @@ def publish(self, settings, item):
...
-
-        # update shot clip xml file with this publish
         try:
+            # update shot clip xml file with this publish
             self._update_flame_clip(item)
-        except Exception, e:
-            raise("Unable to update Flame clip xml: %s" % (e,))
+        except Exception as exc:
+            raise
+            # raise Exception("Unable to update Flame clip xml: %s" % exc)
```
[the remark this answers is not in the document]
thebeeland: Shit. Thanks for catching that.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +135,146 @@ def accept(self, settings, item):
...
+            # the shot, but there might be cases where this heads off a problem.
+            publish_entity_type = get_published_file_entity_type(publisher.sgtk)
+
+            self.logger.debug("Clip publish types to search for: %s" % CLIP_PUBLISH_TYPE)
+
+            clip_publishes = publisher.shotgun.find(
+                publish_entity_type,
+                [
+                    ["entity", "is", item.context.entity],
+                    ["published_file_type.PublishedFileType.code", "is", CLIP_PUBLISH_TYPE],
```
manneohrstrom: If we want to support TankPublishedFiles, we need to update `published_file_type.PublishedFileType.code` to be correct for TankPublishedFiles. My suggestion would be to just support PublishedFiles.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +135,146 @@ def accept(self, settings, item):
...
+                    "published_file_type",
+                    "description",
+                ),
+                order=[dict(field_name="id", direction="desc")],
+            )
+
+            if clip_publishes:
+                self.logger.debug("Found published clips: %s" % clip_publishes)
+
+            for clip_publish in clip_publishes:
```
manneohrstrom: This approach is nice because it is trying to gracefully handle the case when the clip file isn't there. But i wonder if we are over thinking it - why not just do a `find_one()`, get the latest publish and if that doesn't point to a valid file, handle that case. btw i am not feeling strongly about this, i just feel that perhaps it's a little bit simpler and therefore better.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
```diff
@@ -126,66 +135,146 @@ def accept(self, settings, item):
...
+                else:
+                    self.logger.debug(
+                        "Published clip file isn't accessible: %s" % flame_clip_path
+                    )
         else:
             self.logger.debug(
-                "No flame clip template configured. "
-                "Plugin %s no accepting item: %s" % (self, item)
+                "Unable to look up a clip file to update. No template exists, "
+                "and the item's context is not associated with an entity. This "
```
manneohrstrom: nit: the part of the comment speculating about the project context is perhaps not necessary?

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
manneohrstrom: it's my old crappy code, sorry.

hooks/tk-multi-publish2/basic/nuke_update_flame_clip.py
thebeeland: Does that equate to just leaving this how it is? I was thinking we weren't taking TankPublishedFile into consideration anymore.

manneohrstrom: You also need to get rid of `publish_entity_type = get_published_file_entity_type(publisher.sgtk)` and hard code `PublishedFile`.

PR #52, opened 2018-05-09
For SG-4609: Handles "node" type engine commands in Nuke Studio.

thebeeland:
This allows us to have something like tk-nuke-quickdailies configured for Nuke Studio without having it show up in the Shotgun menu on the main menu bar. Instead, we now separate out those engine commands and add them to the Nuke style context menu.


robblau:
Definitely what @thebeeland says. This app is intended to be used as a node and depends on the graph connections to function properly. Having it show up in the menu was an oversight rather than desired behavior. 👍 to go ahead with it.

PR #53, opened 2018-05-29
Do not attempt to regenerate menu items in no display mode

ahroneknf:
Verify that engine.has_ui is True before regenerating menu items when the context changes. 

This issue was discovered when running a Nuke script that performs context changes on a render farm.


thebeeland:
Hi @ahroneknf! We've recently incorporated this fix into tk-nuke, so I'm going to close this out!

PR #55, opened 2018-06-22
in on_context_change() add has_ui check

aparna-d2:
Doing `engine.context_change()` while running a `nuke -t` session produces the error:
> 'NoneType' object has no attribute 'create_menu'

Checking for has_ui before attempting to create menu fixes that.


manneohrstrom:
Great! Thank you for this! I'll fold it into master now and it will be part of the next integrations release.

PR #56, opened 2018-07-06
added ability to set hotkeys in the tk-nuke.yml settings file.

kramester:
added the ability to specify hotkeys in the tk-nuke.yml config file like this:

```yaml
settings.tk-nuke.shot_step:
  menu_favourites:
  - {app_instance: tk-multi-workfiles2, name: File Open..., hotkey: "ctrl+o"}
  - {app_instance: tk-multi-snapshot, name: Snapshot..., hotkey: "alt+shift+s"}
```


manneohrstrom:
thanks! This seems like a really useful change!!
internal note: we should make sure to update the documentation as well as the default configs to include sensible hotkeys.

PR #58, opened 2018-10-10
Startup optimization on registered callbacks

aoblet:
To speed up the startup, I made the following updates on registered callbacks:

If a current engine is running
1. Avoid to create a tank object
1. engine_refresh: try to change_context with the new context instead of starting a new engine each time

### DIFF
--- a/classic_startup/Python/Startup/__init__.py
+++ b/classic_startup/Python/Startup/__init__.py
@@ -61,7 +61,7 @@ def _setup_sgtk():
         return
 
     try:
-        engine = tank.platform.start_engine(engine_name, context.tank, context)
+        tank.platform.start_engine(engine_name, context.tank, context)
     except Exception, e:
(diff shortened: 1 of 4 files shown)


aoblet:
Any opinions on this ?

pscadding:
Hi I apologise, this seems to have slipped under our radar. I've actually created a pull request my self recently to fix a bunch of bugs around the automatic context switching, and as part of that, I moved the registering of the callbacks into the engine rather than having them in the startup code.
https://github.com/shotgunsoftware/tk-nuke/pull/59/files
This means that the engine can now be bootstrapped without needing to handle the registry of callbacks manually.
It also contains a fix similar for the engine context switching, though a bit different.
I'm sorry we didn't see this.

aoblet:
Ok, so I add some remarks on your pull request.

PR #59, opened 2018-11-30
SG-9833 SG-9835 SG-9836 context switching improvements

pscadding:
This addresses three tickets that centre around the automatic context change and callbacks.
The Fixes are:

1. The automatic context switching callbacks are no longer initialized outside of the engine in the startup code. This means you can now bootstrap the nuke engine without having to call this part separately.
2. The automatic context switching now uses the engine.context_change method to switch contexts when an engine is available. This means that the panel app will update its context for example when a user opens or saves a file through Nuke's open menu.
3. The automatic context switching now respects that different environments might have different states of true or false for this setting. In short, the behaviour should now match the other engines.

Other things of note are:

- I updated the code to use the `pre_app_init` engine method rather than the `engine_init` method
- The callback startup logic/behaviour has changed. Previously it would check if there was a file. If not it would gather the context from the environment variable `TANK_NUKE_ENGINE_INIT_CONTEXT`, and if there was a file it would gather the context for that. Now it always boots up in the previous context, loads the engine, then checks if there is a file and if the automatic_context_switch setting is true before trying to match the context of the file. This brings it in line with the other engine's behaviours and means it now properly respects the setting.
- The `pre_app_init_nuke` (previously `engine_init_nuke`) method now gets called at post context change so that the environment variables are appropriately updated. This was necessary because otherwise a new instance of Nuke starting up would use the original boot context rather than the current state before the new instance was launched.

**UPDATED**

This was previously not working with plugin-based bootstrapping, partly due to my changes but it also never properly work either. I've now updated it to work with the plugin bootstrap.

I've also removed all the custom new Nuke session spawn startup, and instead, it now uses the same startup code as if it was launching Nuke for the first time from the `tk-multi-launchapp`. This simplifies things, and according to my test has no negative impact on the behaviour.
Note a bunch of the classic startup scripts have been removed as a result.

**UPDATED 14th Mar 2019**

Added another fix, where Nuke was calling the init.py multiple times, causing it to attempt to bootstrap the engine multiple times. I've added a slightly hacky fix to the init.py to check if its already been called and if so not attempt to bootstrap again. This wasn't a problem previously as we cleaned up the tank env vars so bootstraping a second time wasn't possible, but with the change to standardize the startup and subsequent automatic context change boots, this meant the env vars are still set and it allowed to bootstrap for every time the init.py was called. I've reached out to the Foundry to see if there is a better way to handle this.


python/tk_nuke/__init__.py
```diff
@@ -120,21 +116,23 @@ def __tank_on_save_callback():
         # API instance.
         try:
             tk = tank.tank_from_path(file_name)
```
aoblet: Instead of trying to create a new `tank` object in any case, we could maybe check  if there is a current engine and take the associated `tank` object.

Like [here](https://github.com/shotgunsoftware/tk-nuke/pull/58/files#diff-2082bc816b93dbd02e442c7942fa7709R140)

python/tk_nuke/__init__.py
```diff
@@ -144,56 +142,59 @@ def tank_startup_node_callback():
...
-    
         else:
+            new_ctx = tk.context_empty()
+
+        # now restart the engine with the new context
+        __engine_refresh(tk, new_ctx)
+
+        # If we have opened a file then we should check if automatic context switching is enabled and change if possible
+        engine = tank.platform.current_engine()
+        if nuke.root().name() != "Root" and engine.get_setting("automatic_context_switch"):
```
aoblet: We could maybe store `nuke.root().name()` result into a variable since we use it later on.

python/tk_nuke/__init__.py
```diff
@@ -144,56 +142,59 @@ def tank_startup_node_callback():
...
             # file->open
             file_name = nuke.root().name()
-            
+
             try:
                 tk = tank.tank_from_path(file_name)
```
aoblet: We can maybe fetch the `tank` object from the current engine instead of creating it in each case, what do you think ?

python/tk_nuke/__init__.py
```diff
@@ -144,56 +142,59 @@ def tank_startup_node_callback():
...
-                except:
-                    new_ctx = tk.context_empty()
-            else:
+
+        # first load up in our previous context so we can start the engine an work out if automatic context switching
+        # is enabled, and if it is we should then switch to the correct environment.
+        # base it on the context we 'inherited' from the prev session
+        # get the context from the previous session
+        project_root = os.environ.get("TANK_NUKE_ENGINE_INIT_PROJECT_ROOT")
+        tk = tank.Tank(project_root)
```
aoblet: We can maybe fetch the `tank` object from the current engine instead of creating it in each case, what do you think ?

python/tk_nuke/__init__.py
```diff
@@ -82,27 +82,23 @@ def __create_tank_error_menu():
     
 def __engine_refresh(tk, new_context):
```
aoblet: Since `new_context` has a `tank` object as a member, we could use it directly instead of passing as an extra argument, no ?

Like [here](https://github.com/shotgunsoftware/tk-nuke/pull/58/files#diff-2082bc816b93dbd02e442c7942fa7709R88)

python/tk_nuke/__init__.py
pscadding: Thanks for the suggestions. I've spent a little bit of time looking at this, but I don't think we can reliably reuse the sgtk instance. If the new script file path is in a different project, then this will cause errors later.
If you do the following:
`sgtk.platform.current_engine().sgtk.context_from_path(file_name)`
it won't error (unlike `tk = tank.tank_from_path(file_name)`), but will return back a context containing only the site url, and then it will attempt to start the engine in site context, where as really it should be disabling the Shotgun integration.

engine.py
```diff
@@ -144,12 +144,20 @@ def in_plugin_mode(self):
...
+        Sets up the engine into an operational state. This method called before
+        any apps are loaded.
         """
+
         self.logger.debug("%s: Initializing...", self)
 
+        local_python_path = os.path.abspath(os.path.join(self.disk_location, "python"))
+        # append the local path so that the tk_nuke and tk_nuke_qt packages can be imported
+        sys.path.append(local_python_path)
+        import tk_nuke
```
thebeeland: Is it possible to use `self.import_module` here instead of having to tinker with sys.path?

engine.py
```diff
@@ -559,6 +554,15 @@ def post_context_change(self, old_context, new_context):
         :param old_context: The sgtk.context.Context being switched away from.
         :param new_context: The sgtk.context.Context being switched to.
         """
+        # As we've changed contexts, we should update our environment variables so that if we spawn a new nuke instance
+        # it will start up in the same environment.
+        self.pre_app_init_nuke()
+
+        # Make sure the callbacks are updated based one the current environment settings.
+        # (for example they may be enabled or disabled in the new environment.)
+        import tk_nuke
```
thebeeland: Same here: maybe `self.import_module` will work?

plugins/basic/Python/tk_nuke_basic/plugin_bootstrap.py
```diff
@@ -298,14 +302,37 @@ def _report(self, progress_value, message):
...
         # At this point we are guaranteed that bootstrapped is finished, so we can dismiss progress
         # reporting.
         self._progress_task.done()
 
+        # We should attempt to switch to the context that was supplied from the Nuke session that launched/spawned this
+        # session if we can. If this Nuke session has just been launched via the launch app this env var won't be set.
+        if not failed and self._previous_ctx_str:
+            # We have a context set previously, this must mean that this Nuke instance has been spawned from
+            # another which had TK context. We should switch to this context.
+            self._logger.debug("Context found from env var: \"TANK_CONTEXT\", context: %s" %  self._previous_ctx_str)
```
thebeeland: [minor] Instead of concatenating the string yourself, you can let the log handler do it like below. That will keep Python from having to concat the string unless debug logging is turned on. In this case that's not going to be a performance issue, but in situation where there's a debug log message that's very high frequency, having to concat the string when it's not actually going to be logged (ie: debug logging is off) would be wasteful.

`self._logger.debug("Some message: %s that came from %s", (my_message, my_location))`

python/tk_nuke/__init__.py
```diff
@@ -119,86 +117,93 @@ def __tank_on_save_callback():
...
-        
+
 g_tank_callbacks_registered = False
 
-def tank_ensure_callbacks_registered():   
+
+def tank_ensure_callbacks_registered(engine=None):
     """
     Make sure that we have callbacks tracking context state changes.
+    The OnScriptLoad callback really only come into play when your opening a file or creating a new script, when there is
```
thebeeland: "you're" ;)

pscadding: and come[s]!

engine.py
pscadding: Yeah it seems import_module is built for that really, however I see this comment further down:

> \# Note! not using the import as this confuses Nuke's callback system
> \# (several of the key scene callbacks are in the main init file).

https://github.com/shotgunsoftware/tk-nuke/blob/master/engine.py#L295-L296
So perhaps there is a good reason for not doing this?

pscadding: seems like that warning should probably be added as a comment in these other locations as well?

pscadding: Actually, I tested this and it did, in fact, break it, integrations no longer loaded, and without error as well.

PR #62, opened 2019-03-22
SG-11138: Improved callback logging

manneohrstrom:
Adds extensive debug and exception trap logging to nuke callbacks.


python/tk_nuke/__init__.py
```diff
@@ -70,123 +74,153 @@ def __create_tank_error_menu():
...
+
         else:
             # file->open
             file_name = nuke.root().name()
-            
+            logger.debug("File->Open detected with file name '%s'" % file_name)
+
             try:
                 tk = tank.tank_from_path(file_name)
+                logger.debug("Tk instance '%r' associated with path '%s'" % (tk, file_name))
```
pscadding: Jeff picked me up on this so I'm going to raise it here :D. Small thing but it can save some performance here when debug is not enabled by switching this line to the following:
`logger.debug("Tk instance '%r' associated with path '%s'" tk, file_name)`
So that the string only gets built when if it gets logged.

pscadding:
This looks great, will certainly help debug any issues around this code. I'm expecting that there will be a number of changes however when this comes to be merged due to the recent changes in the automatic context switching.

PR #67, opened 2019-12-11
SG-14164: Resolves shotgunpanel crashing in Nuke 12.

thebeeland:
A small qss tweak in the shotgunpanel app resolved the crashing problems on OSX, but the issue persisted despite that on Windows and Linux. That crash in Nuke 12 does not appear to be tied to the qss itself, but the fact that ANY qss is being applied at the exact moment that it is in our panel-specific logic in tk-nuke.

This fix works and has no visible impact for the user. We're just deferring the application of the qss until Qt has a chance to finish whatever it's doing at that moment.

### DIFF
--- a/python/tk_nuke_qt/panels.py
+++ b/python/tk_nuke_qt/panels.py
@@ -212,8 +212,25 @@ class ToolkitWidgetWrapper(QtGui.QWidget):
             bundle.logger.debug("Created new toolkit panel widget %s", self.toolkit_widget)
             
             # now let the core apply any external stylesheets
-            bundle.engine._apply_external_styleshet(bundle, self.toolkit_widget)
-            
+            #
+            # NOTE: To be honest, we're not entirely sure why this is required. In Nuke 12
+            # we started experiencing a crash when the shotgunpanel app was being launched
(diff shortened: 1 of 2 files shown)


python/tk_nuke_qt/panels.py
```diff
@@ -212,8 +212,25 @@ def __init__(self):
...
+            # gives Qt time process other events before getting to this call. With that in
+            # mind, we did attempt to just make a call to processEvents here to try to get
+            # the same result without a timer, but that did not stop the crash problem.
+            def _set_qss():
+                bundle.engine._apply_external_styleshet(bundle, self.toolkit_widget)
+
+            self._timer = QtCore.QTimer(self.toolkit_widget)
+            self._timer.setSingleShot(True)
+            self._timer.timeout.connect(_set_qss)
+            self._timer.start(1)
```
jfboismenu: Have you tried setting the timeout to 0? Per Qt's documentation:

> A QTimer with a timeout interval of 0 will time out as soon as all the events in the window system's event queue have been processed.

which sounds exactly like what we're attempting to do here.

thebeeland: 👍

PR #68, opened 2020-01-10
Fix context change on save

kramester:
- on_load and on_save callbacks were refreshing all the engine apps every time the user saved/loaded even if the context was unchanged.
- added if statement to prevent refresh if ctx was unchanged


kramester:
This was causing some serious lag every time an artist would hit save.

pscadding:
Hi @kramester, thanks a lot for submitting this. Someone else has submitted something similar before, and when I tested it, it caused issues.

It introduces a bug, where the sgtk menu remains disabled in the scenario where you were in a context (say shot 10, comp), then save file outside of the structure so that the menu becomes disabled. If you then save the file again in the same as the previous context, the menu stays disabled.
This is because when the menu is disabled, it doesn't actually destroy the engine or change context, and the menu only gets rebuilt if you start the engine. So if the disabled menu is present and your context doesn't change when you save a file into the correct location then it will stay disabled.
I appreciate this is a bit of an edge case, and in theory, there is nothing wrong with your suggestion per se but additional work would need to be done to find a way around this.

kramester:
@pscadding, sounds like we just need to store the menu state and add that as a condition on whether to run the callbacks.

The lag on save is the biggest issue here. Artists don't expect to have to wait a couple seconds for a quick CRTL+S save. Adding this condition made the saves instantaneous as they should be. The edge case you're describing is very much an edge case for us as we never save outside the pipeline. We're going to continue to use my patch here until you guys come up with something better.

pscadding:
@kramester Yeah if the refreshing of the context is causing pain, then we should find a way to avoid the refresh. I've logged an internal ticket to look at this.

PR #71, opened 2020-05-26
Black

pscadding:
Prep work for Python 3 port.


README.md
```diff
@@ -1,4 +1,6 @@
-[![Build Status](https://secure.travis-ci.org/shotgunsoftware/tk-nuke.png?branch=master)](http://travis-ci.org/shotgunsoftware/tk-nuke)
+[![Build Status](https://dev.azure.com/shotgun-ecosystem/Toolkit/_apis/build/status/Engines/tk-nuke?branchName=master)](https://dev.azure.com/shotgun-ecosystem/Toolkit/_build/latest?definitionId=83&branchName=master)
```
pscadding, line 1: It looks like there was a build badge previously. I'm not actually sure if I'm supposed to replace the old one. @jfboismenu what should I do here?

jfboismenu: We can have both on for now. Travis runs the tests right now. When we port to Python 3 Azure will run everything and we can remove travis then.

PR #72, opened 2020-05-26
SG-17693 fix to allow nuke to launch on py3

pscadding:
This isn't a whole Python 3 port of the Nuke engine, that still needs to happen when a Python 3 build of Nuke is available.

However this fix does allow Nuke (Py2) to be launched from a Python 3 environment using the `tk-multi-launchapp`.


pscadding:
The reason I think it didn't work in Python 3 is because you used to get the value `linux2` I think, and now it is just `linux`. I agree it doesn't look as nice, but method is there so we don't have to hard code any platform names in our code.

PR #73, opened 2020-09-23
Handle Hiero API deprecated projectRoot methods

dealmeida:
In Nuke 12, some methods of the API `hiero.core` are deprecated and will be removed. That's the case for `Project.projectRoot()` and `Project.setProjectRoot()`

This Pull Request proposes to handle these cases with wrapper methods that will call the API method recommended by Hiero 
Documentation when the Nuke version being used is superior or equal to 12.

https://learn.foundry.com/hiero/developers/12.0/hieropythondevguide/api/api_core.html

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -764,8 +764,8 @@ class NukeEngine(tank.platform.Engine):
         """
         import hiero
 
-        # The nested methods below are wrapper functions that call the appropriate hiero project_root
-        # functions according to the nuke version. Some methods are deprecated in Nuke 12.
+        # Since some Hiero projectRoot methods are deprecated in Nuke 12, we use the wrapper methods
+        # below to call the appropriate Hiero methods according to the Nuke version being used.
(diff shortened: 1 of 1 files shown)


aoblet:
@pscadding, @jfboismenu what do you think ?

jfboismenu:
Hi!
Does this mean that our Hiero integration is currently broken with Nuke 12? I find the wording from Foundry confusing. The methods are listed as part of the API but being removed?

dealmeida:
Hello,

In **12.1** - the only minor version we've tested until now - the integration is not broken: there's this warning printed on terminal (_[DEPRECATION WARNING] Project.setProjectRoot() will be removed in Nuke 12..._) but the methods still work. We haven't tested the **12.2** version.

Would you like us to check with The Foundry if they plan to actually remove them in a specific minor version of Nuke 12?

dealmeida:
Hi,
I've just tested the same methods - `projectRoot()` and `setProjectRoot()` - with **nuke 12.2v1** and they both work in spite of the warning messages.

PR #75, opened 2020-10-30
SG-19954: Fixes a freeze when launching Nuke v11.3 with debugging logging on.

jfboismenu:
NEW FIX DESCRIPTION: The run_at_startup commands are executed while the splashscreen is visited, which is when nuke informs us that the GUI is ready. Unfortunately, it appears that calling async_execute_to_main_thread will lock up at that point of the startup process. I've tried reimplementing it in tk-core with `QMetaObject`, just like  `execute_from_main_thread`, but it still locked up. This is unexpected as a `Qt.QueueConnection` is expected to be non-blocking. What I suspect is happening here is that the main thread has a lock the list of events in the main thread and is not relinquishing it for some reason and our background thread is stuck waiting for the main thread to relinquish it so it can send it's signal.

In the end, I ended up doing something else: postponing the launch of startup commands until after the main thread is done executing events by using `QTimer.singleShot`.

ORIGINAL FIX DESCRIPTION: Originally, we passed print and msg in directly and it locked up Nuke v11.3 when it happened. There's probably some weird issue when passing a CPython function instead of a pure Python method to a Qt signal, so wrap the call into a lambda to make the issue go away.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -619,7 +619,12 @@ class NukeEngine(tank.platform.Engine):
                 nuke.warning("Shotgun Warning: %s" % msg)
 
         # Sends the message to the script editor.
-        self.async_execute_in_main_thread(print, msg)
+        # Originally, this line passed print and msg in directly.
+        # and it locked up Nuke v11.3 when it happened. There's probably
+        # some weird issue when passing a CPython function instead of
+        # a pure Python method to a Qt signal, so wrap the call into a
+        # lambda to make the issue go away.
+        self.async_execute_in_main_thread(lambda: print(msg))
 
     #####################################################################################
     # Panel Support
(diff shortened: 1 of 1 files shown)


toolkitty (approved):

eshokrgozar (approved):

engine.py
```diff
@@ -521,13 +521,23 @@ def _run_commands_at_startup(self):
...
-        # FIXME: This pattern is horrible.
-        setattr(tank, "_callback_from_non_pane_menu", True)
-        try:
-            for command in commands_to_run:
-                command()
-        finally:
-            delattr(tank, "_callback_from_non_pane_menu")
+        def run_at_startup():
+            # FIXME: This pattern is horrible.
+            tank._callback_from_non_pane_menu = True
```
jfboismenu, line 526: I change this line because the dog asked me to. Otherwise it's essentially the same code as before, but called from the single shot timer.

eshokrgozar (approved):

PR #79, opened 2021-02-04
SG-37311 Update engine.py

reformstudios:
It's probably bad practice to permanently change hiero's log level just to print a sg debug log.
HieroPlayer doesn't have a frameserver, so the warning in this file fires every second : /apps/foundry/nuke/Nuke12.2v4/pythonextensions/site-packages/hiero/ui/FnStatusBar.py
To prevent this, we need to revert hiero's log level to what it was before we emit a debug level message.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -611,6 +611,8 @@ class NukeEngine(tank.platform.Engine):
         if self.hiero_enabled:
             import hiero
 
+            existing_log_level = hiero.core.log.logLevel()
+
             if record.levelno >= logging.ERROR:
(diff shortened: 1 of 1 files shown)


tk421storm:
why hasn't this been accepted yet? seeing this same error in our studio, spamming DEBUG logs into the script editor. We know that printing into the script editor can cause QT-related freezes as well

eyeteajay:
This is also spamming the logs at CBS. Please review and fix.

carlos-villavicencio-adsk (approved):

PR #95, opened 2022-11-01
SG-25374 Weblogin does not show up in Nuke 11 and makes Nuke 12 and 13 to crash.

000paradox000:
When ShotGrid session expires and Nuke is open, if you try open a toolkit dialog like save file, Nukes crashes trying to show  web login dialog (exec_ method).

This happens silently, so there's not exception to trap. 

This PR adds a messagebox that warns user that ShotGrid integration won't work until you restart the DCC.

image

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -251,6 +251,69 @@ class NukeEngine(sgtk.platform.Engine):
         os.environ["TANK_ENGINE"] = self.instance_name
         os.environ["TANK_CONTEXT"] = sgtk.context.serialize(self.context)
 
+        """
+        https://jira.autodesk.com/browse/SG-25374
+        Weblogin does not show up in Nuke 11 and makes Nuke 12 and 13 to crash
+        
+        To avoid Nuke crash, a monkeypatch of on_dialog_closed is required,
+        here the user is warned about restarted nuke is needed to continue. 
+        """
+        sgtk.authentication.sso_saml2.core.sso_saml2_core.SsoSaml2Core.on_dialog_closed = self._sso_on_dialog_closed_monkeypatch  # noqa
+
+    @staticmethod
+    def _on_dialog_closed_monkeypatch(self, result):
+        """
+        Called whenever the dialog is dismissed.
+
+        This can be the result of a callback, a timeout or user interaction.
+
+        :param result: Qt result following the closing of the dialog.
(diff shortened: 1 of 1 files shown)


engine.py
```diff
@@ -251,6 +251,70 @@ def pre_app_init_nuke(self):
...
         os.environ["TANK_CONTEXT"] = sgtk.context.serialize(self.context)
 
+        """
+        https://jira.autodesk.com/browse/SG-25374
+        Weblogin does not show up in Nuke 11 and makes Nuke 12 and 13 to crash
+
+        To avoid Nuke crash, a monkeypatch of on_dialog_closed is required,
+        here the user is warned about restarted nuke is needed to continue.
+        """
+        sgtk.authentication.sso_saml2.core.sso_saml2_core.SsoSaml2Core.on_dialog_closed = (
```
patrick-hubert-adsk, line 261: This works ?  I was expecting you to need to use `setattr()`

If it works, great !

patrick-hubert-adsk (approved):
LGTM.
Much better approach to modify the Nuke engine.

engine.py
000paradox000: Yes it works well.

shaynacohen:
Hi @000paradox000 , can this read "Your ShotGrid session has expired and you will need to restart Nuke."

shaynacohen (changes requested):
I'll need some clarity on the workflow/functionality so that I can provide some more suggestions.

pamintabi:
Pasting notes from [the Slack thread](https://autodesk.slack.com/archives/C87SBC2KH/p1667305424083519) here for reference:

Couple of notes to improve the experience (consistency between SG in Nuke and SG Web and to provide the user with more info and with actionable options):

- Instead of the red circle with X icon, could we use the same icon in SG? I believe it's the triangle or the diamond caution icon

- For the copywriting: maybe something like` The ShotGrid session has expired, to continue using ShotGrid in Nuke, please restart Nuke`

- Since Restart cannot be actioned by a button in this dialog, it's fine to leave the `OK` button as is

000paradox000:
Changes applied:

- New icon
- New message

image

shaynacohen (approved):
This looks good to me!

PR #97, opened 2023-04-25
Add setting for keeping Project context if couldn't resolve sgtk from path (SG-30710)

RicardoMusch:
Currently, if someone opens a file with Nuke's File --> Open, that does not reside in any known project path, the project context is lost and the ShotGrid menu is disabled.

This pull request adds an engine setting that allows for keeping at least the project context enabled so that the user can use the shotgrid workfiles app to save the current file into the pipeline.
There are various workflows where this is preferable over simply disabling toolkit.

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -19,6 +19,12 @@ configuration:
                      context every time the currently loaded file changes. Defaults to True."
         default_value: True
 
+    allow_keep_context_from_project:
+        type: bool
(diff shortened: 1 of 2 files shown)


NorberMV:
Thank you very much for this contribution @RicardoMusch ! , we have opened this [PR](https://github.com/shotgunsoftware/tk-nuke/pull/99) which is the branch where we are working, and we have included these changes.

PR #99, opened 2023-07-18
SG-30710 Add Setting For Keeping Project Context If Could Not Resolve sgtk From Path

NorberMV:
This originally comes from this [PR ](https://github.com/shotgunsoftware/tk-nuke/pull/97), and aims to introduce a new engine setting that avoids the ShotGrid menu from being disabled when for example someone opens a file with Nuke's File --> Open, that does not reside in any known project path, allowing users to retain the Project context enabled so that the user can use the shotgrid workfiles app to save the current file into the pipeline.
Fixes #97

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -19,6 +19,12 @@ configuration:
                      context every time the currently loaded file changes. Defaults to True."
         default_value: True
 
+    allow_keep_context_from_project:
+        type: bool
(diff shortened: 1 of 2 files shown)


carlos-villavicencio-adsk (approved):
LGTM

PR #119, opened 2025-07-16
SG-39698 Remove try/except sgutil blocks and ensure_ methods

eduardoChaucaGallegos:
### **Description**
- Removed try/except sgutils block

### DIFF
--- a/python/tk_nuke/menu_generation.py
+++ b/python/tk_nuke/menu_generation.py
@@ -19,11 +19,6 @@ import traceback
 import nukescripts.openurl
 import nukescripts
 
-try:
-    from tank_vendor import sgutils
-except ImportError:
-    from tank_vendor import six as sgutils
-
 # -----------------------------------------------------------------------------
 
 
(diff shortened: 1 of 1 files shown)


python/tk_nuke/menu_generation.py
```diff
@@ -829,8 +824,11 @@ def get_documentation_url_str(self):
             doc_url = self.app.documentation_url
             # Deal with nuke's inability to handle unicode.
             if type(doc_url) == str:
-                doc_url = sgutils.ensure_str(
-                    unicodedata.normalize("NFKD", doc_url), "ascii", "ignore"
+                doc_url = unicodedata.normalize("NFKD", doc_url)
+                doc_url = (
+                    doc_url.decode("ascii", "ignore")
+                    if isinstance(doc_url, bytes)
+                    else str(doc_url)
```
julien-lang, line 831: Given the history of this method and why it was created, I suggest to clean it up:


```suggestion
if not self.app:
  return

return self.app.documentation_url
```

PR #124, opened 2025-10-07
SG-43889 Add a setting to suppress the minimum supported Nuke version Dialog

RicardoMusch:
Add a setting to suppress the minimum supported Nuke version Dialog.
This is required for usage in a Studio pipeline where we are aware of the "danger" of running into issues but should be able to suppress these dialogs for our users.

**Reasons:**
- Minimum supported is not minimum required version so we can still use the engine but can expect bugs.
- It will become very annoying for users if Nuke blocks every time they load a new instance, something a comper might do a hundred times a day.
- This gives Pipeline TD's the ability to turn off the GUI Warning (it still logs the warning)


python/tk_nuke/__init__.py
```diff
@@ -156,11 +156,53 @@ def __sgtk_on_save_callback():
...
+        __ensure_sgtk_tab_and_context()
+
+        # save the context into the sgtk_context knob
+        root = nuke.root()
+        if "sgtk_context" in root.knobs():
+            context_knob = root.knob("sgtk_context")
+            context_str = new_ctx.serialize(with_user_credentials=False, use_json=False)
+            context_knob.setValue(context_str)
+            logger.debug("Updated 'sgtk_context' knob with value: %s" % context_str)
+
```
julien-lang, line 169: Hi @RicardoMusch, I don't understand how the changes in this file are related to this PR. Could you clarify this, please?

PR #126, opened 2025-10-23
SG-40319 Tests nuke versions below 10

eduardoChaucaGallegos:
### **Description**
This pull request updates the Nuke version tests in `tests/test_startup.py` by removing the `@skip` decorators from several test methods. This change re-enables tests for Nuke versions below 10, which were previously skipped due to string comparison issues related to the removal of `distutils`.

Testing improvements:

* Re-enabled tests for Nuke versions 9, 8, 7 (including 7.0v9), and 6 by removing `@skip` decorators from their respective test methods in `tests/test_startup.py`. This allows these tests to run again and ensures coverage for older Nuke versions.

### **Dependencies**
This PR fixes the underlying issue that was causing the version comparison failures for Nuke versions below 10. The root cause was improper version string parsing in tk-core's `version_parse` function, which was falling back to alphabetical string comparison instead of semantic version comparison.

**⚠️ IMPORTANT: This PR should NOT be merged until the following dependency is approved and merged:**

- [x] **tk-core PR**: Add `_normalize_version_format()` function to properly handle Nuke version format (X.YvZ) in `version_parse()` - [[Link to tk-core PR](https://github.com/shotgunsoftware/tk-core/pull/1066)]
- [x] Verify that tk-core changes are deployed and available.
- [x] Update azure-pipelines.yml from [this current PR](https://github.com/shotgunsoftware/tk-nuke/pull/126/files#diff-7915b9b726a397ae7ba6af7b9703633d21c031ebf21682f3ee7e6a4ec52837a5L17) with the pointing to master.
- [x] Re-run tests to confirm version filtering works correctly with the new tk-core version.

### **Technical Details**
The issue was that Nuke versions like "6.3v6", "10.0v5", "12.2v5" couldn't be parsed by `packaging.version.parse()` due to the "vX" format, causing fallback to string comparison where "13.0v1" > "6.3v6" evaluates to `False` (alphabetically). The tk-core fix normalizes these to standard semantic versioning format ("6.3.6", "10.0.5", "12.2.5") before parsing.

### DIFF
--- a/startup.py
+++ b/startup.py
@@ -38,7 +38,7 @@ class NukeLauncher(SoftwareLauncher):
         "major_minor_version": r"[\d.]+",
     }
 
-    # Templates for all the display names of the products supported by Nuke 6, 7 and 8.
+    # Templates for all the display names of the products supported by Nuke 7 and 8.
     NUKE_7_8_PRODUCTS = [
         "Nuke",
         "NukeX",
--- a/startup.py
+++ b/startup.py
@@ -38,8 +38,8 @@ class NukeLauncher(SoftwareLauncher):
         "major_minor_version": r"[\d.]+",
     }
 
-    # Templates for all the display names of the products supported by Nuke 7 and 8.
-    NUKE_7_8_PRODUCTS = [
+    # Templates for all the display names of the products supported by Nuke 6, 7 and 8.
+    NUKE_6_7_8_PRODUCTS = [
         "Nuke",
         "NukeX",
         "NukeAssist",
@@ -203,8 +203,8 @@ class NukeLauncher(SoftwareLauncher):
         """
         # As of Nuke 6, Nuke versions formatting is <Major>.<Minor>v<patch>.
         # This will grab the major version.
-        if version.split(".", 1)[0] in ["7", "8"]:
-            return self.NUKE_7_8_PRODUCTS
+        if version.split(".", 1)[0] in ["6", "7", "8"]:
+            return self.NUKE_6_7_8_PRODUCTS
         else:
             return self.NUKE_9_OR_HIGHER_PRODUCTS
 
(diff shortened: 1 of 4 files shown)


julien-lang (approved):

tests/test_startup.py
```diff
@@ -341,50 +335,48 @@ def test_nuke10(self):
...
         """
         Ensures we are returning the right variants for Nuke 8.
         """
-        self._test_nuke([], "8.0v4")
+        self._test_nuke(
+            [
+                "Nuke 8.0v4",
+                "NukeX 8.0v4",
+                "NukeAssist 8.0v4",
+            ],
```
carlos-villavicencio-adsk: As far as I understand, this list should be empty because we're not supporting Nuke 8. Same with the other unsupported versions.

startup.py
```diff
@@ -38,8 +38,8 @@ class NukeLauncher(SoftwareLauncher):
         "major_minor_version": r"[\d.]+",
     }
 
-    # Templates for all the display names of the products supported by Nuke 7 and 8.
-    NUKE_7_8_PRODUCTS = [
+    # Templates for all the display names of the products supported by Nuke 6, 7 and 8.
+    NUKE_6_7_8_PRODUCTS = [
```
carlos-villavicencio-adsk: Currently, we support Nuke > 14. It's ok to leave a version lower than 10 just for testing purposes, but I don't think it's relevant to test starting from 6 each version of Nuke.

carlos-villavicencio-adsk (commented):
I don't think this PR addresses the issue here. Usually, they were empty lists `[]` because those versions are unsupported. Last time I investigated, the issue was caused by the version parsing mechanism, but I don't remember if this was originated in this engine or in multi-launchapp.

julien-lang (approved):

carlos-villavicencio-adsk (approved):

PR #134, opened 2026-06-22
SG-43461 SG-43644 Migrate Flow host support for Nuke

yungsiow:
Implement subclass of FlowHost for Nuke (including dependency introspection support)
Instantiate NukeHost instance during engine set up with context is configured with flow integration

### DIFF
--- a/python/flowam/host.py
+++ b/python/flowam/host.py
@@ -1,0 +1,654 @@
+# -
+# *****************************************************************************
+# Copyright 2026 Autodesk, Inc. All rights reserved.
+#
+# These coded instructions, statements, and computer programs contain
+# unpublished proprietary information written by Autodesk, Inc. and are
+# protected by Federal copyright law. They may not be disclosed to third
+# parties or copied or duplicated in any form, in whole or in part, without
+# the prior written consent of Autodesk, Inc.
+# *****************************************************************************
+#
+
+import os
+
+from tank import LogManager
+from tank.flowam.host import FlowHost
+from tank.flowam.utils import search_file_expression
+from tank_vendor.flow_integration_sdk.dependency import DependencyData
+from tank_vendor.flow_integration_sdk.utils import (
+    cleanpath,
+    fileext,
+    trace,
+)
+
+import nuke
+
+
+class NukeHost(FlowHost):
+    """Nuke implementation of FlowHost interface.
+    This is a collection of required capabilities to support Flow AM integration.
+    """
+
+    logger = LogManager.get_logger("NukeHost")
+
+    #: Desktop application deals with generic workfiles
+    WORKFILE_TYPE = "type.workfile.nuke"
+    #: Nuke native file types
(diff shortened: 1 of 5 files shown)


python/flowam/host.py
```diff
@@ -0,0 +1,654 @@
...
+# unpublished proprietary information written by Autodesk, Inc. and are
+# protected by Federal copyright law. They may not be disclosed to third
+# parties or copied or duplicated in any form, in whole or in part, without
+# the prior written consent of Autodesk, Inc.
+# *****************************************************************************
+#
+
+import os
+
+from tank import LogManager
```
carlos-villavicencio-adsk, line 15: We prefer importing `sgtk` instead of `tank`. I know it's a dynamic alias but that's how we're doing it for the last years.

carlos-villavicencio-adsk (commented):
Just a small suggestion. The rest looks good overall. I didn't checked logic because it's out of my knowledge.

python/flowam/host.py
yungsiow: Actually I tried to use sgtk first and it caused import issues.  Only tank works which is why I went with that.  I guess it has to do with when the module is first access which is very early in the engine's set up process.

carlos-villavicencio-adsk (approved):
LGTM

python/flowam/host.py
```diff
@@ -0,0 +1,654 @@
...
+
+
+class NukeHost(FlowHost):
+    """Nuke implementation of FlowHost interface.
+    This is a collection of required capabilities to support Flow AM integration.
+    """
+
+    logger = LogManager.get_logger("NukeHost")
+
+    #: Desktop application deals with generic workfiles
```
chenm1adsk: Please update this stale comment

python/flowam/host.py
```diff
@@ -0,0 +1,654 @@
...
+
+        Args:
+            text: Text to be copied.
+
+        Returns:
+            True on success.
+        """
+        from tank.platform.qt import QtGui as qtg
+
+        qtg.QApplication.instance().clipboard().setText(text)
```
chenm1adsk: Should we consider add
```
app = QtGui.QApplication.instance()
        if app is None:
```
for all engine like tk-desktop incase they are launched headless with no GUI? like from tk-shell that Julien mentioned?

chenm1adsk (approved):
Left a question and a small comment. The rest looks good to me

python/flowam/host.py
yungsiow: So I don't think tk-shell launches engines headless.  It's just a command line way of launching them, but they launch in the exact same way I believe.  Having said that, I'll put the check in to be safe.  Thanks!

yungsiow: done!

python/flowam/host.py
yungsiow: fixed!

PR #135, opened 2026-06-29
SG-43811 migrate flow write node

yungsiow:
**Summary:**  Port the FlowWrite node implementation in Nuke of POC to Nuke engine, and adapt the user flow more in line with existing SG Write node.

**Details:**
- Added FlowWriteNode class to Nuke engine
- Kept original create and pre-render callbacks attached to node to ensure node is created under the right conditions, and renders go to the right place
- Removed the post-render callback that published renders to Flow - this will be handled via the Publisher app from now on
- Ensured the Nuke collector detects FlowWrite node renders and types them correctly, storing pertinent information in publish item about the FlowWrite renders
- Add a publish script for FlowWrite nodes specifically which will do some post processing on the node after the publish is complete
- During NukeHost initialization, register the FlowWrite node in the tab menu of Nuke and hook up callbacks

### DIFF
--- a/hooks/tk-multi-publish2/basic/collector.py
+++ b/hooks/tk-multi-publish2/basic/collector.py
@@ -12,6 +12,8 @@ import os
 import nuke
 import sgtk
 
+from flowam.flow_write_node.flow_write_node import FlowWriteNode
+
 HookBaseClass = sgtk.get_hook_baseclass()
 
 # A look up of node types to parameters for finding outputs to publish
@@ -98,6 +100,7 @@ class NukeSessionCollector(HookBaseClass):
         # run node collection if not in hiero
         if hasattr(engine, "hiero_enabled") and not engine.hiero_enabled:
             self.collect_sg_writenodes(project_item)
+            self.collect_flow_writenodes(project_item)
             self.collect_node_outputs(project_item)
 
     def collect_current_nuke_session(self, settings, parent_item):
@@ -257,6 +260,11 @@ class NukeSessionCollector(HookBaseClass):
             # iterate over each instance
             for node in all_nodes_of_type:
 
+                # Skip FlowWrite nodes here
+                # We will handle them explicitly in collect_flow_writenodes()
+                if FlowWriteNode.is_flow_write(node):
+                    continue
+
                 param_name = _NUKE_OUTPUTS[node_type]
 
                 # evaluate the output path parameter which may include frame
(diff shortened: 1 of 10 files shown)


hooks/tk-multi-publish2/basic/nuke_flow_write_publish_script.py
```diff
@@ -0,0 +1,130 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+
+from tank_vendor.flow_integration_sdk.objects import FlowAsset
+
+HookBaseClass = sgtk.get_hook_baseclass()
+
+
+class NukeFlowWritePublishPlugin(HookBaseClass):
```
chenm1adsk, line 16: Sorry I am not quite understand nuke hooks here. I am wondering why we need two different publish hooks? Are they get triggered differently?

python/flowam/host.py
```diff
@@ -82,6 +87,25 @@ def __init__(self, context):
...
+        # Add FlowWrite node to toolbar
+        self.logger.info("Adding toolbar commands...")
+        nuke.toolbar("Nodes").addCommand(
+            "Flow/FlowWrite",
+            FlowWriteNode,  # callback to instantiate a write node
+            icon="Write.png",
+        )
+
+        self.logger.info("Registering global callbacks for Write nodes...")
+        nuke.addKnobChanged(FlowWriteNode._on_update_flow_write, nodeClass="Write")
```
chenm1adsk, line 100: Do we need to remove registered callback on tear down? Would this registered callback accumulate?

python/flowam/flow_write_node/flow_write_node.py
```diff
@@ -0,0 +1,403 @@
...
+        """
+        # Get current node context
+        write_node = nuke.thisNode()
+
+        if not cls.is_flow_write(write_node):
+            return  # This is not a flow write node
+
+        # Check for mandatory properties
+        if not write_node["asset_name"].value().strip():
+            nuke.message("Asset Name must be set.")
```
chenm1adsk, line 319: I am wondering why we need two messages here. Does nuke.message display differently?

chenm1adsk (commented):
Left some questions :)

hooks/tk-multi-publish2/basic/nuke_flow_write_publish_script.py
yungsiow: Yes, and my understanding is also not great.  The original one only deals with Nuke "session" publishes - i.e. nuke script files.  The item filter only accepts these.  And since I need specific publish behaviour for the renders (which is different than scripts) - specifically a bit of post processing - the way to ensure I can "catch" those publishes is by creating a custom publisher which does the normal generic publish, but then some extra functionality after.  Hope that makes sense!

python/flowam/host.py
yungsiow: Yes this is a good idea.  I haven't been doing that in the host code and probably should.  Let me look into that!

python/flowam/flow_write_node/flow_write_node.py
yungsiow: Yes I think Nuke.message() pops up a little dialog.

python/flowam/host.py
yungsiow: Addressed!

hooks/tk-multi-publish2/basic/nuke_flow_write_publish_script.py
chenm1adsk: Thanks for the explanation. I am thinking should we move nuke_flow_write_publish_script.py inside nuke_publish_script.py and use context to check if we trigger flowam logic for nuke script publish workflow?
In this case, user don't need to update their config to add a new hook to be able to use flowam logic (same as collector.py/pre_publish.py/publish_file.py in native tk-multi-publsih2).
What do you think?

python/flowam/flow_write_node/callbacks.py
```diff
@@ -0,0 +1,250 @@
...
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Autodesk, Inc.
+
+"""
+The FlowWrite node is designed to be pipeline/integration agnostic.
+This modules contains the callbacks that support our specific Flow integration.
+"""
+
+from __future__ import annotations  # needed for python 3.9 support
```
carlos-villavicencio-adsk: Since this code is not going to run on Houdini (because this is Nuke engine) probably we don't need this. Not sure if I'm missing something.

python/flowam/flow_write_node/callbacks.py
```diff
@@ -0,0 +1,250 @@
...
+        write_node: Nuke FlowWrite node that triggered callback.
+
+    Raises:
+        RuntimeError
+    """
+    import sgtk
+    from .flow_write_node import FlowWriteNode
+
+    # Ensure flow host exists
+    host = sgtk.platform.current_engine().flow_host
```
carlos-villavicencio-adsk, line 128: Do we want to do a core check here as well?

carlos-villavicencio-adsk (commented):
I believe the code style validation check if failing.

chenm1adsk (approved):
LGTM!

python/flowam/flow_write_node/callbacks.py
yungsiow: Is it the case that we don't support any Nuke engines that run python 3.9?  I just never know when we need this and when we don't!

carlos-villavicencio-adsk: Yes, Nuke 15 is the min [version supported](https://help.autodesk.com/view/SGDEV/ENU/?guid=SGD_si_integrations_engine_supported_versions_html) and it comes with Python 3.10.

hooks/tk-multi-publish2/basic/collector.py
```diff
@@ -406,6 +485,10 @@ def _get_node_colorspace(self, node):
             cs = cs[9:-1]
         return cs
 
+    def _in_flow_context(self) -> bool:
+        """Return True if in a Flow-enabled context."""
+        return hasattr(self.parent.engine.context, "flow_project_id")
```
carlos-villavicencio-adsk:
```suggestion
        return getattr(self.parent.engine.context, "flow_project_id", None) is not None
```

python/flowam/flow_write_node/callbacks.py
yungsiow: Ok will remove!

python/flowam/flow_write_node/callbacks.py
yungsiow: In this case, I don't think so.  If you have a FlowWrite node already created, you should be in a flow context.

hooks/tk-multi-publish2/basic/collector.py
yungsiow: Updated!

